PR #12, opened 2022-05-09
unquote path string from LSP

aihuasxy:
File path from LSP maybe contain quoted char, for example '@' quoted to '%40'


manateelazycat:
你 rebase 一下补丁吧。

PR #18, opened 2022-05-09
Add javascript/typescript support

Bitnut:


VitalyAnkh (commented):
I think there shouldn't be hardcoded paths...

PR #24, opened 2022-05-10
Fix for Windows

vconcat:
For now, at least `gopls` works.


manateelazycat:
@vconcat Let me know if this PR can merge, thanks for your work!!! ;)

manateelazycat:
Thanks for patch.

But I fix issuse cause by this patch : https://github.com/manateelazycat/lsp-bridge/commit/4ef0cad3ae32b01a38eb7f1d6efa2ba438b8fb66

We should add cwd in pipe arguments, because some language (such as python) can works without project folder, then project_path is equal python filepath.

PR #37, opened 2022-05-11
fix protocol exchange problem to some lsp servers under Windows

wsw0108:
like pyright/rust-analyzer/ts-ls

### DIFF
--- a/core/lspserver.py
+++ b/core/lspserver.py
@@ -23,6 +23,7 @@ from sys import flags, stderr
 from core.utils import path_as_key, generate_request_id, path_to_uri, uri_to_path
 from subprocess import PIPE
 from threading import Thread
+import io
 import json
 import os
 import queue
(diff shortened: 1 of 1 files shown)


manateelazycat:
这个补丁什么原理？能否描述一下，没看懂这个补丁

wsw0108:
1. subprocess.Popen的text=True，实际也是用io.TextIOWrapper包裹了binary形式打开的文件描述符
2. https://docs.python.org/3/library/io.html#io.TextIOWrapper
> When reading input from the stream, if newline is None, [universal newlines](https://docs.python.org/3/glossary.html#term-universal-newlines) mode is enabled. Lines in the input can end in '\n', '\r', or '\r\n', and these are translated into '\n' before being returned to the caller. If newline is '', universal newlines mode is enabled, but line endings are returned to the caller untranslated. If newline has any of the other legal values, input lines are only terminated by the given string, and the line ending is returned to the caller untranslated.

如果用io.TextIOWrapper包裹时，指定了newline=''，则对消息中的\r\n不过转换。

manateelazycat:
> 1. subprocess.Popen的text=True，实际也是用io.TextIOWrapper包裹了binary形式打开的文件描述符
> 2. https://docs.python.org/3/library/io.html#io.TextIOWrapper
> 
> > When reading input from the stream, if newline is None, [universal newlines](https://docs.python.org/3/glossary.html#term-universal-newlines) mode is enabled. Lines in the input can end in '\n', '\r', or '\r\n', and these are translated into '\n' before being returned to the caller. If newline is '', universal newlines mode is enabled, but line endings are returned to the caller untranslated. If newline has any of the other legal values, input lines are only terminated by the given string, and the line ending is returned to the caller untranslated.
> 
> > When writing output to the stream, if newline is None, any '\n' characters written are translated to the system default line separator, [os.linesep](https://docs.python.org/3/library/os.html#os.linesep). If newline is '' or '\n', no translation takes place. If newline is any of the other legal values, any '\n' characters written are translated to the given string.
> 
> 如果用io.TextIOWrapper包裹时，指定了newline=''，则向stdin写消息时，对\r\n中的\n不做转换。

感谢补丁

PR #44, opened 2022-05-12
fix: `length=` introduced at Emacs version 28.1

zy9306:
length= not work in emacs27


zerolfx:
Could you also change the Emacs version requirements in lines 14 and 15?

zy9306:
> Could you also change the Emacs version requirements in lines 14 and 15?

Changed

PR #57, opened 2022-05-13
Handle `textDocument/hover` response according to lsp spec

vritser:
`textDocument/hover` spec: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_hover


manateelazycat:
Cool

PR #65, opened 2022-05-14
Beautify the document

vritser:
QQ20220514-144454@2x


manateelazycat:
Did we need add https://github.com/jrblevin/markdown-mode in lsp-bridge dependences?

vritser:
Yes, we need

PR #96, opened 2022-05-17
Fix find define key error

xinyifly:
Currently when I use find def in `elixir` , `*lsp-bridge*` reports:
```
--- Send (56637): textDocument/definition

--- Recv response (56637): textDocument/definition
Error when processing response 56637
  File "/home/xinyifly/.config/emacs/site-lisp/lsp-bridge/core/handler/__init__.py", line 63, in handle_response
    self.process_response(response)
  File "/home/xinyifly/.config/emacs/site-lisp/lsp-bridge/core/handler/find_define.py", line 18, in process_response
    file_info = response[0]
KeyError: 0
```

Inspecting `response` I got:
```
{'range': {'end': {'character': 6, 'line': 1}, 'start': {'character': 6, 'line': 1}}, 'uri': 'file:///home/xinyifly/git/lino/device/lib/lino_device/gateway.ex'}
```

Which is a dict instead of a list of dict, this pull request handles that situation.

### DIFF
--- a/core/handler/find_define.py
+++ b/core/handler/find_define.py
@@ -1,5 +1,6 @@
 from core.handler import Handler
 from core.utils import *
+from typing import Union
 
 
 class FindDefine(Handler):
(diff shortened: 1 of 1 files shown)


manateelazycat:
Thanks

PR #105, opened 2022-05-18
Make snippet work with dart_analysis_server

50ways2sayhard:


lsp-bridge.el
```diff
@@ -547,8 +547,7 @@ Then LSP-Bridge will start by gdb, please send new issue with `*lsp-bridge*' buf
                (if (minibufferp) (window-buffer (minibuffer-selected-window))
                  (current-buffer))
 
-             (let ((snippet-fn (and (string= kind "Snippet")
```
manateelazycat, line 550: Why delete (string= kind "Snippet") ?

manateelazycat (changes requested):

lsp-bridge.el
manateelazycat: But if you delete this line, snippet-fn always is true, we can't do lsp-bridge--apply-text-edits if Kind is auto-import.

50ways2sayhard: I see. I'll push a commit to fix it.

50ways2sayhard: @manateelazycat done

PR #133, opened 2022-05-23
fix disable and reenable corfu auto completion

ctrl2wei:
need remove corfu--auto-post-command from post-command-hook if manually enable lsp-bridge-mode
and corfu auto completion already enabled.

### DIFF
--- a/lsp-bridge.el
+++ b/lsp-bridge.el
@@ -967,6 +967,8 @@ If optional MARKER, return a marker instead"
 
        (setq-local corfu-auto-prefix 0)
        (setq-local corfu-auto nil)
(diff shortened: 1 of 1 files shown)


lsp-bridge.el
```diff
@@ -967,6 +967,8 @@ If optional MARKER, return a marker instead"
 
        (setq-local corfu-auto-prefix 0)
        (setq-local corfu-auto nil)
+       (remove-hook 'post-command-hook #'corfu--auto-post-command 'local)
```
manateelazycat, line 970: 为什么需要这一行，这一行解决了什么问题呢？

PR #142, opened 2022-05-26
Correct the spelling of the command to install fonts

haidong:
I think there is a minor typo. Thanks for the wonderful work!


manateelazycat:
Thanks for help.

PR #154, opened 2022-05-28
[WIP] Add support for xref

theFool32:
Hacky code that sleeps and checks for the variable to turn the asynchronous find-definition/find-references into a synchronous one.

Still WIP since many boundary conditions are not considered. Need help from the community.


manateelazycat:
无法接受 sleep-for 的方案，建议在 lsp-bridge 外部自己写一个函数包装 lsp-bridge-jump-to-def 和 xref 函数。

抱歉无法合并。

PR #193, opened 2022-06-08
Hide yas completion after "."

lyjdwh:


manateelazycat:
Thanks for patch, but we should hardcode `.` in lsp-bridge.

I have push https://github.com/manateelazycat/lsp-bridge/commit/cf9f12b6fee93d22ad302311ba1aa6a10bed7a05 to fixed this.

PR #212, opened 2022-06-14
:sparkles: support yaml

zwpaper:
1. support `yaml`
2. fix wrong `lsp-bridge-lang-server-list` to `lsp-bridge-lang-server-mode-list`
3. add file name in `add support language` docs


manateelazycat:
感谢补丁

PR #230, opened 2022-06-22
ACM popup, support for buffer centering modes

adouzzy:
For buffer centering modes, the acm popup at the wrong place

image

Also there is a typo.


manateelazycat:
截图 2022-06-22 12-29-51

It's wrong for buffer not centering postion,  I have revert this patch.

manateelazycat:
https://github.com/manateelazycat/lsp-bridge/commit/a80b6b9279a03ea98313cc6a7b345f897cfc73cd has fixed this issue

We should use `window-body-width` instead.

PR #234, opened 2022-06-23
support zig-mode

zhscn:


manateelazycat:
thanks

manateelazycat:
Can you add zig info to README?

Thanks

PR #263, opened 2022-07-09
Add Fortran language support

nujabse:
Add lsp support for `f90-mode`, `fortran-mode` for Fortran language using `fortls` language server.


manateelazycat:
thanks

PR #264, opened 2022-07-09
Fix document hover not displaying under some language servers

ssl19:
Some language server, for example jdtls, don't have `range` key in
textDocument/hover response. According to
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/,
t, `range` is optional in the response of `Hover Request`.


manateelazycat:
Thanks for patch.

PR #265, opened 2022-07-09
Use substitude-command-keys in lsp-bridge-ref-update-header-line

ssl19:
Use `substitude-command-keys` instead of hard coded keybinding indicator
in header-line of *lsp-bridge-ref* buffer


manateelazycat:
Cool

PR #304, opened 2022-08-10
fix: clojure mode keyword complete

vinurs:


manateelazycat:
Right way should https://github.com/manateelazycat/lsp-bridge/commit/2b2e36d35ceb540d6005959fbefccb3ad6b1e1d1

PR #329, opened 2022-08-29
fix: add default mode hook for `erlang-mode`

JimMoen:


JimMoen:
`lsp-bridge` won't start by default before this commit.

PR #338, opened 2022-09-02
require dash in tabnine-bridge

falood:
https://github.com/manateelazycat/lsp-bridge/blob/master/acm/tabnine-bridge.el#L257-L263

need dash for the functions `--remove` and `--filter` here


MatthewZMD:
Thanks!

manateelazycat:
I think better way is remove dash functions, its not necessary

PR #360, opened 2022-09-21
feat: use python-epc for tabnine

theFool32:
elisp <-> python <-> tabnine

The codes are directly used from my repo and just work. Give it a try to check whether the performance is satisfactory :)
However, the codes need to be optimized in the future, e.g., no need for another file of *-epc.el.


acm/tabnine/utils.py
```diff
@@ -0,0 +1,114 @@
...
+        sys_architecture = arch_translations[sys_architecture]
+
+    if sysinfo.system == "Windows":
+        sys_platform = "pc-windows-gnu"
+
+    elif sysinfo.system == "Darwin":
+        sys_platform = "apple-darwin"
+
+    elif sysinfo.system == "Linux":
+        sys_platform = "unknown-linux-musl"
```
manateelazycat, line 74: 要改成 unknown-linux-gnu

manateelazycat (changes requested):

manateelazycat:
这个补丁在 Linux 下还没法用， 暂时不能合并， 各位开发者请知晓。

manateelazycat:
Fixed by https://github.com/manateelazycat/lsp-bridge/commit/b5f472e2185f80a2d0f1b6688ba5a1d55713a5c5

PR #367, opened 2022-09-28
fix: Don't check the last command before fetching doc

twlz0ne:
Check the last command in `acm-fetch-candidate-doc` causes the doc to sometimes not be fetched.

## Steps to reproduce

1. Switch to buffer `*scratch*`.
2. Type `(global-m`.
3. Press `M-j` to scroll doc.
4. Press `M-n` to select next candidate.

## Expected Behavior

Doc frame would be updated.

## Actual Behavior

Doc frame not updated.


manateelazycat:
Looks good, thanks.

PR #377, opened 2022-10-11
Fix issue that sometimes can't get documentation

twlz0ne:
* acm/acm-backend-lsp.el (acm-backend-lsp-candidate-doc): Don't fetch the documentation if it is not empty.
* lsp-bridge.el (lsp-bridge-fetch-completion-item-info): Replace `label` with `key` in the judgment condition, make the logic consistent with `FileAction.completion_item_resolve`.
(lsp-bridge-completion--record-items): Reset fetch tick


manateelazycat:
Thanks patch.

PR #384, opened 2022-10-21
escape char in prefix before build regexp

ssl19:
Support elisp symbol containing special characters, like "foo+bar", "foo?bar" and "foo*bar"


manateelazycat:
Thanks for patch.

PR #393, opened 2022-10-30
Add Phpactor as PHP's Language Server

zonuexe:
[Phpactor](https://github.com/phpactor/phpactor) is an implementation of PHP's Language Server.

### DIFF
--- a/lsp-bridge.el
+++ b/lsp-bridge.el
@@ -323,6 +323,11 @@ Then LSP-Bridge will start by gdb, please send new issue with `*lsp-bridge*' buf
   "Default LSP server for C language, you can choose `clangd' or `ccls'."
   :type 'string)
 
+(defcustom lsp-bridge-php-lsp-server "intelephense"
+  "Default LSP server for PHP language, you can choose `intelephense' or `phpactor'."
+  :type 'string
+  :safe #'stringp)
+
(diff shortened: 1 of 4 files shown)


manateelazycat:
Can you add lsp-bridge-php-lsp-server like lsp-bridge-python-lsp-server does, user can customize lsp-bridge-php-lsp-server to choose PHP LSP server.

lsp-bridge.el
```diff
@@ -323,6 +323,11 @@ Then LSP-Bridge will start by gdb, please send new issue with `*lsp-bridge*' buf
   "Default LSP server for C language, you can choose `clangd' or `ccls'."
   :type 'string)
 
+(defcustom lsp-bridge-php-lsp-server "intelephense"
+  "Default LSP server for PHP language, you can choose `intelephense' or `phpactor'."
+  :type 'string
+  :safe #'stringp)
```
zonuexe, line 329: I want to switch servers in `.dir-locals.el` per project in PHP, so I mark it as a [safe variable](https://www.gnu.org/software/emacs/manual/html_node/emacs/Safe-File-Variables.html) by setting `:safe`.

zonuexe:
@manateelazycat Thank you for reviewing. I added custom variable and rebased and squashed this branch.

manateelazycat:
Awesome!

PR #394, opened 2022-10-30
Add .editorconfig

zonuexe:
[EditorConfig](https://editorconfig.org/) is useful for defining file indentation style. I usually write JSON with 4 spaces, so I think this setting is important.


manateelazycat:
Thanks patch, but not everyone use EditorConfig, we can add .editorconfig in .gitignore, then you can add .editorconfig in lsp-bridge for your need.

PR #395, opened 2022-10-30
Use default-directory when load-file-name is NIL (eval-buffer)

zonuexe:
I often try to reload with `M-x eval-buffer` in development, so it makes sense to fallback to `default-directory` when `load-file-name` is undefined


manateelazycat:
Thanks for patch.

PR #399, opened 2022-11-02
fix: repair lsp-volar config

androidvov:
repair lsp-volar config


manateelazycat:
@androidvov 

Did you got error `Request initialize failed with message: Cannot read properties of undefined (reading ’replace’)` when start volar lsp server?

PR #406, opened 2022-11-08
Proposal: Activate lsp-bridge-render-markdown-content on acm-doc

zonuexe:
Language servers return the doc as Markdown, so it's convenient to render it.

## Before

スクリーンショット 2022-11-08 21 46 10

## After

スクリーンショット 2022-11-08 21 57 43

PR #440, opened 2022-12-01
Apply markdown styling to acm doc, similar to lsp-bridge

JordanAnthonyKing:
This adds similar styling of the acm doc popup to the styling in lsp-bridge.

Note: commented out lines are lines that would depend on variables defined in lsp-bridge. These can be required or deleted.

### DIFF
--- a/acm/acm.el
+++ b/acm/acm.el
@@ -908,11 +908,7 @@ The key of candidate will change between two LSP results."
               (insert (if (stringp candidate-doc)
                           candidate-doc
                         (format "%S" candidate-doc)))
-	      ;; (markdown-mode)
-	      ;; (message (eval markdown-hide-markup))
-              ;; (visual-line-mode 1)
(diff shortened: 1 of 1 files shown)


manateelazycat:
Thanks for path, this patch same as https://github.com/manateelazycat/lsp-bridge/pull/406

We can't active lsp-bridge-render-markdown-content in acm-doc, this patch will greatly slow down lspbridge's completion speed.

Sorry, lsp-bridge's first design target is speed, if some feature slow down lsp-bridge's performance, we will disable those feature until it fix performance issue.

manateelazycat:
@JordanAnthonyKing  I have push new commit https://github.com/manateelazycat/lsp-bridge/commit/dc6e03dcd2e1e50636bb696f137dc3c92831f1ab

Can you try it? 

New patch only render markdown content when idle 200ms, this patch won't slow down lsp-bridge's completion speed.

PR #443, opened 2022-12-02
clojure-lsp: turn on support single file

suliveevil:
ref: https://github.com/clojure-lsp/clojure-lsp/issues/1402#issuecomment-1335098967


suliveevil:
turn on single file support to avoid warning when open single edn file.

manateelazycat:
clojure确实不支持单文件啊，单文件的时候没法补全

PR #449, opened 2022-12-04
fix: acm-cancel-timer define to late

kongds:
Macro `acm-cancel-timer` should define before using.
Defining to late will cause error `invalid function acm-cancel-timer` when we load native compiled lsp-bridge.


manateelazycat:
Thanks buddy. ;)

PR #452, opened 2022-12-04
Proposal: acm-backend-custom.el for user extension

zonuexe:
acm is not extensible right now, but what about adding interface like this?


manateelazycat:
Did you have any completion backend request that acm-mode haven't ?

I personally hold conservative opinions on adding custom completion interfaces, because most Emacsers don't understand multi-thread, if it copy some capf code for customarily completion backend, it will block lsp-bridge.

We should push more async completion backends to acm directory.

PR #453, opened 2022-12-04
Use bound-and-true-p instead of symbol-value

zonuexe:
This is a minor fix: `bound-and-true-p` simplifies the code slightly.

There is no change in behavior in this case, but `bound-and-true-p` behaves more intuitively in most cases. This is because `symbol-value` does not work as expected for lexically bound variables.

### DIFF
--- a/acm/acm-backend-lsp.el
+++ b/acm/acm-backend-lsp.el
@@ -230,8 +230,7 @@ If optional MARKER, return a marker instead"
 (defun acm-backend-lsp-snippet-expansion-fn ()
   "Compute a function to expand snippets.
 Doubles as an indicator of snippet support."
-  (and (boundp 'yas-minor-mode)
-       (symbol-value 'yas-minor-mode)
+  (and (bound-and-true-p 'yas-minor-mode)
        'yas-expand-snippet))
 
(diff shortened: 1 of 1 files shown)


acm/acm-backend-lsp.el
```diff
@@ -230,8 +230,7 @@ If optional MARKER, return a marker instead"
 (defun acm-backend-lsp-snippet-expansion-fn ()
   "Compute a function to expand snippets.
 Doubles as an indicator of snippet support."
-  (and (boundp 'yas-minor-mode)
-       (symbol-value 'yas-minor-mode)
+  (and (bound-and-true-p 'yas-minor-mode)
```
manateelazycat, line 233: Change to 

```elisp
(and (boundp 'yas-minor-mode) 
         (symbol-value 'yas-minor-mode)
         yas-minor-mode))
```

Otherwise Github CI will failed, I guess Emacs in CI haven't define function `bound-and-true-p`

manateelazycat (changes requested):

manateelazycat:
Fixed by https://github.com/manateelazycat/lsp-bridge/commit/437c0c823ead30d08ff88b030afc179b9abe119c

Thanks for patch.

PR #491, opened 2023-01-01
feat: Allow popup completion menu for string interpolation

naoking158:
Thank you for developing this wonderful package.

I propose new feature to allow popup completion menu for string interpolation.
This only works with string interpolation.

For example in python-mode,

no popup menu in string,
image

popup menu in string interpolation,
image

### DIFF
--- a/lsp-bridge.el
+++ b/lsp-bridge.el
@@ -519,6 +519,13 @@ you can customize `lsp-bridge-get-workspace-folder' to return workspace folder p
     (default                    . standard-indent)) ; default fallback
   "A mapping from `major-mode' to its indent variable.")
 
+(defcustom lsp-bridge-string-interpolation-open-chars-alist
+  '((python-mode . "[^\$]\{")
+    (yaml-mode . "\{\{")
+    (t . "\$\{"))
+  "Open characters for string interpolation. The elements are cons cell (major-mode . open-char-regexp)"
+  :type 'cons)
+
 (defun lsp-bridge--get-indent-width (mode)
   "Get indentation offset for MODE."
   (or (alist-get mode lsp-bridge-formatting-indent-alist)
@@ -933,6 +940,17 @@ So we build this macro to restore postion after code format."
   (not (or (member lsp-bridge-last-change-command lsp-bridge-completion-stop-commands)
            (member (format "%s" last-command) lsp-bridge-completion-stop-commands))))
 
+(defun lsp-bridge-string-interpolation-p (string-interpolation-open-chars-alist)
+  "Check if the cursor position is subject to string interpolation"
+  (save-excursion
+    (let* ((initial-pos (point))
+           (search-char (cdr (or (assoc (buffer-local-value 'major-mode (current-buffer))
+                                        string-interpolation-open-chars-alist)
+                                 (assoc t string-interpolation-open-chars-alist))))
+           (open-pos (search-backward-regexp search-char nil t)))
+      (if open-pos
+          (not (search-forward-regexp "\}" initial-pos t))))))
+
 (defun lsp-bridge-not-in-string ()
   "Hide completion if cursor in string area."
   (or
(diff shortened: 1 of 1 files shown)


lsp-bridge.el
```diff
@@ -519,6 +519,13 @@ you can customize `lsp-bridge-get-workspace-folder' to return workspace folder p
     (default                    . standard-indent)) ; default fallback
   "A mapping from `major-mode' to its indent variable.")
 
+(defcustom lsp-bridge-string-interpolation-open-chars-alist
+  '((python-mode . "[^\$]\{")
+    (yaml-mode . "\{\{")
+    (t . "\$\{"))
```
naoking158, line 525: The completion menu may popup in unintended situations if this part is not adjusted.

manateelazycat: I think we should remove `(t . "\$\{"))`, if some langugage need, then we add new rule.

lsp-bridge.el
```diff
@@ -933,6 +940,17 @@ So we build this macro to restore postion after code format."
...
            (member (format "%s" last-command) lsp-bridge-completion-stop-commands))))
 
+(defun lsp-bridge-string-interpolation-p (string-interpolation-open-chars-alist)
+  "Check if the cursor position is subject to string interpolation"
+  (save-excursion
+    (let* ((initial-pos (point))
+           (search-char (cdr (or (assoc (buffer-local-value 'major-mode (current-buffer))
+                                        string-interpolation-open-chars-alist)
+                                 (assoc t string-interpolation-open-chars-alist))))
+           (open-pos (search-backward-regexp search-char nil t)))
```
manateelazycat, line 950: You need use `save-excursion` wrap `search-backward-regexp` outside, then you can remove initial-pos, it's unnecessary.

manateelazycat (changes requested):

lsp-bridge.el
```diff
@@ -933,6 +940,17 @@ So we build this macro to restore postion after code format."
   (not (or (member lsp-bridge-last-change-command lsp-bridge-completion-stop-commands)
            (member (format "%s" last-command) lsp-bridge-completion-stop-commands))))
 
+(defun lsp-bridge-string-interpolation-p (string-interpolation-open-chars-alist)
+  "Check if the cursor position is subject to string interpolation"
+  (save-excursion
+    (let* ((initial-pos (point))
+           (search-char (cdr (or (assoc (buffer-local-value 'major-mode (current-buffer))
```
manateelazycat, line 947: If `search-char` is nil, we should stop this function.

Becasue we should remove `(t . "\$\{"))` from `lsp-bridge-string-interpolation-open-chars-alist`, then `search-char` should be nil if current mode is not python-mode.

manateelazycat (changes requested):

lsp-bridge.el
naoking158: Fixed in [151cb8f](https://github.com/manateelazycat/lsp-bridge/pull/491/commits/151cb8fcfb09b583719eb3bc5e0b3d8f09d87d99)

lsp-bridge.el
naoking158: Fixed in [151cb8f](https://github.com/manateelazycat/lsp-bridge/pull/491/commits/151cb8fcfb09b583719eb3bc5e0b3d8f09d87d99)

naoking158:
Thank you for your kind and thoughtful comments.
I fixed.

manateelazycat:
awesome patch

manateelazycat:
I have push new commit https://github.com/manateelazycat/lsp-bridge/commit/d932c67725e1549554973f34d0813e88f9540490 to simpler patch. ;)

naoking158:
Thanks a lot!!!

BTW, I noticed you are the author of the EAF.
I am also a fan of EAF.
Thanks for the great package!!!

PR #526, opened 2023-02-07
Use `_` (wildcard) instead of `t` for pcase

zonuexe:
```
lsp-bridge/acm/acm.el: Warning: Pattern t is deprecated.  Use `_' instead
```

See https://www.gnu.org/software/emacs/manual/html_node/elisp/pcase-Macro.html

https://github.com/emacsmirror/emacs/blob/013655811aa1c89754372610c8c6ccccec166035/lisp/emacs-lisp/pcase.el#L948-L951


manateelazycat:
thanks for patch

PR #557, opened 2023-03-06
Support byte compilation

kira-bruneau:
Hi! I've been trying to [package this extension with Nix](https://github.com/kira-bruneau/home-config/blob/d7e88d71f076867dbef3a2b3d5f66e167d949624/package/emacs/package/lsp-bridge/default.nix) and I ran into some problems with byte compilation:

```
acm/acm-backend-yas.el:188:10: Error: ‘add-to-list’ can’t use lexical var ‘candidates’; use ‘push’ or ‘cl-pushnew’
acm/acm-backend-tempel.el:24:16: Error: ‘add-to-list’ can’t use lexical var ‘candidates’; use ‘push’ or ‘cl-pushnew’
acm/acm-backend-elisp.el:112:10: Error: ‘add-to-list’ can’t use lexical var ‘candidates’; use ‘push’ or ‘cl-pushnew’
acm/acm-backend-telega.el:52:18: Error: ‘add-to-list’ can’t use lexical var ‘candiates’; use ‘push’ or ‘cl-pushnew’
acm/acm-backend-search-file-words.el:101:16: Error: ‘add-to-list’ can’t use lexical var ‘candidates’; use ‘push’ or ‘cl-pushnew’
acm/acm-backend-tailwind.el:95:14: Error: ‘add-to-list’ can’t use lexical var ‘candidates’; use ‘push’ or ‘cl-pushnew’
acm/acm-backend-citre.el:195:16: Error: ‘add-to-list’ can’t use lexical var ‘candidates’; use ‘push’ or ‘cl-pushnew’
acm/acm.el:93:2: Error: Cannot open load file: No such file or directory, acm-icon
lsp-bridge-ref.el:585:8: Error: ‘add-to-list’ can’t use lexical var ‘file-extensions’; use ‘push’ or ‘cl-pushnew’
lsp-bridge-code-action.el:83:2: Error: Cannot open load file: No such file or directory, acm
lsp-bridge-diagnostic.el:83:2: Error: Cannot open load file: No such file or directory, acm
lsp-bridge.el:87:2: Error: Cannot open load file: No such file or directory, acm-frame
lsp-bridge-call-hierarchy.el:2:2: Error: Cannot open load file: No such file or directory, acm-frame
```

For the `add-to-list` errors I replaced `add-to-list` with `push`, `mapcon`, `mapcan`, and `seq-filter`, which got it working! A nice upside to this is now these loops run in $O(n)$ instead of $O(n^2)$ - I doubt the change in performance will be noticeable though :sweat_smile:

For the acm errors I pulled out the code that automatically adds acm to `load-path`, and updated all references to explicitly require adding acm to `load-path`. I think that's the right approach, but it's another step to add to your config, and I'm curious what you think.

I also thought it would be a good idea to add a new test to verify byte compilation. There are still warnings, but this PR gets through all the errors!

After getting this all set up, I've been really liking this package! Thanks for all the hard work! It _really_ makes a difference how you've designed things around a fast feedback loop 😊

### DIFF
--- a/acm/acm-backend-citre.el
+++ b/acm/acm-backend-citre.el
@@ -168,40 +168,43 @@
 
 (defun acm-backend-citre-candidates (keyword)
   (when acm-enable-citre
-    (let* ((candidates (list)))
-      (if (string-empty-p keyword)
-          (progn
-            (setq-local acm-backend-citre-prefix-keyword 
-                        (save-excursion
-                          (backward-char)
-                          (if (string= (char-to-string (following-char)) ".")
-                              (thing-at-point 'symbol)
-                            "")))
-            ;; When there is nothing in front of the . , it will get nil
-            (unless acm-backend-citre-prefix-keyword
-              (setq-local acm-backend-citre-prefix-keyword ""))
-            (setq-local acm-backend-citre-search-keyword acm-backend-citre-prefix-keyword))
-        (if (string-suffix-p (substring keyword 0 (1- (length keyword))) acm-backend-citre-search-keyword)
-            (setq-local acm-backend-citre-search-keyword (concat acm-backend-citre-search-keyword (substring (string-reverse keyword) 0 1)))
-          (setq-local acm-backend-citre-search-keyword keyword)))
-      (let* ((collection (unless (string-empty-p acm-backend-citre-search-keyword)
-                           (let ((cands (citre-get-completions)))
-                             (delete-dups (mapcar #'citre-capf--make-candidate (nth 2 cands)))))))
-        (when acm-backend-citre-keyword-complete
-          (setq collection (append (acm-backend-citre-get-mode-keyword) collection)))
-        (when collection
-          (dolist (candidate collection)
+    (if (string-empty-p keyword)
+        (progn
+          (setq-local acm-backend-citre-prefix-keyword
+                      (save-excursion
+                        (backward-char)
+                        (if (string= (char-to-string (following-char)) ".")
+                            (thing-at-point 'symbol)
+                          "")))
+          ;; When there is nothing in front of the . , it will get nil
+          (unless acm-backend-citre-prefix-keyword
+            (setq-local acm-backend-citre-prefix-keyword ""))
+          (setq-local acm-backend-citre-search-keyword acm-backend-citre-prefix-keyword))
+      (if (string-suffix-p (substring keyword 0 (1- (length keyword))) acm-backend-citre-search-keyword)
+          (setq-local acm-backend-citre-search-keyword (concat acm-backend-citre-search-keyword (substring (string-reverse keyword) 0 1)))
+        (setq-local acm-backend-citre-search-keyword keyword)))
+    (let* ((collection (unless (string-empty-p acm-backend-citre-search-keyword)
+                         (let ((cands (citre-get-completions)))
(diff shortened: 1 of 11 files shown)


manateelazycat:
Thanks for patch, I need some local test before merge it. ;)

lsp-bridge.el
```diff
@@ -80,10 +80,6 @@
 (require 'markdown-mode)
 (require 'diff)
 
-(defvar acm-library-path (expand-file-name "acm" (if load-file-name
-                                                     (file-name-directory load-file-name)
-                                                   default-directory)))
-(add-to-list 'load-path acm-library-path t)
```
manateelazycat, line 86: I don't understand why remove those code? 

I think that making such modifications will increase the cost of using lsp-bridge for users.

kira-bruneau: Oh it's because byte compilation won't automatically pick this up, so I thought it would be better to be explicit all the time. Do you you think it would be better to leave in though?

kira-bruneau: Oh yep it probably won't, but the main reason I wanted to get this working is that the Nix packages expect the files to be byte compiled, and as nice bonus it also acts as a lint check.

I think that native compilation will make more a difference (and that's also part of the Nix build process if you have it enabled), and it'll also need an explicit load path for acm.

kira-bruneau:
Hmm I wonder why CI is failing trying to find yasnippet - it should be automatically installed by the init script :thinking:

lsp-bridge.el
manateelazycat: Users can manually load acm into the load-path, but I don't think this code should be removed. Otherwise, all existing lsp-bridge users would be unable to use lsp-bridge after the update, and adding some redundancy would not cause the byte-compilation process to fail.

kira-bruneau:
Ok fixed the CI failures and brought back to the call that automatically loads acm!

manateelazycat:
Thank you for your hard work, the patch is great, you're beautiful. ;) 

Thank you.

manateelazycat:
BTW, you can try my other package: 

https://github.com/manateelazycat/blink-search/
https://github.com/emacs-eaf/emacs-application-framework

EAF and blink-search both use similar multi-threading technology as lsp-bridge.

kira-bruneau:
Oh wow! I'll definitely check out blink search! I really like [my setup with fd + fzf](https://gitlab.com/kira-bruneau/home-config/-/blob/main/package/emacs/config/setup/minor-modes/projectile.el#L64-72), but it looks like this does a lot more than just search through files - and would give would probably give the same kind of responsiveness that running fzf directly in a terminal gives. Before this I used [fzf.el](https://github.com/bling/fzf.el), but I found running a full terminal emulator in a fake "minibuffer" to be way too buggy.

I also didn't realize how big EAF is! I noticed you mention it the EmacsConf2022 video for lsp-bridge, but wow there are so many interesting sub-projects there that I need to check out!

kira-bruneau:
I'm also really curious about the multi-threaded git client! magit is just too slow on some repos like nixpkgs.

manateelazycat:
> I'm also really curious about the multi-threaded git client! magit is just too slow on some repos like nixpkgs.

"Eaf-git can instantly open even the largest projects, such as the Linux kernel, with instant opening and real-time search for even 1 million patches."

MatthewZMD:
Great work!! @kira-bruneau

PR #567, opened 2023-03-18
Add elixir-ts-mode-hook to lsp-bridge-default-mode-hooks and to interpolation list

wkirschbaum:
Sorry, missed this one. I tried adding heex-ts-mode, but the completion was a bit unpleasant as it returns mostly elixir, not heex suggestions.


wkirschbaum:
converting to draft for now, because it seems there are some more tweaks we need to do.

wkirschbaum:
okay, think that is all for now.

manateelazycat:
Awesome!

PR #576, opened 2023-03-25
some improvements for org babel

kongds:
1. support `lsp-bridge-org-babel-lang-list` in lsp-bridge

2. remove begin_src and end_src when check org babel border

PR #577, opened 2023-03-26
use post-affiliated as bop for org babel

kongds:
Use post-affiliated instead of begin to support code block with #+name https://github.com/manateelazycat/lsp-bridge/pull/576#discussion_r1148573154.


manateelazycat:
maybe we need split those code to lsp-bridge-org.el ?

PR #595, opened 2023-04-14
feat: add verilog's verible language server

VitalyAnkh:
Support the [newly announced](https://www.chipsalliance.org/news/integrating-language-server-protocol-in-verible/) verilog language server: [verible](https://github.com/chipsalliance/verible).


manateelazycat:
Thanks man

PR #602, opened 2023-04-20
support lexical for elixir language

falood:


manateelazycat:
Patch looks good, thanks! ;)

PR #629, opened 2023-05-12
Send response after lsp workspace/applyEdit eval in emacs.

zhoupen9:
Fix lsp server (i.e. gopls ) not respond requests issue after lsp-bridge-code-action. Send { "applied": True } to lsp server after received workspace/applyEdit request and eval in emacs.


manateelazycat:
感谢补丁

PR #640, opened 2023-05-24
Save match data

braineo:
~~It is broken again, do not merge plz~~

I wrap the function in before and after-change-functions with `save-match-data`, this should fix the problem of modifying match results. Let me know if you want me to properly indent the code, I was trying to make a minimum change here.


manateelazycat:
Thanks for patch

PR #656, opened 2023-06-17
Add fallback functions for searching def/ref

norris-young:
It will be possible to intergrate lsp-bridge with other packages. 
In my case, I will try to find with tags backend when lsp failed to find def/ref.


manateelazycat:
Cool

Issue #659, opened 2023-06-22
rename-file-advisor interferes with all rename-file


ahmed-shariff:
The `lsp-bridge--rename-file-advisor` function assumes any call to `rename-file` is related to the currently open file. I have https://github.com/Fanael/persistent-scratch running on my system, which has a call to `rename-file` with respect to the scratch-buffers. As a result, the above mentioned function renames the current-buffers file-name as well.

PR #660, opened 2023-06-22
Fix for rename-file-advisor renaming current file accidentally

ahmed-shariff:
This is a fix for #659

### DIFF
--- a/lsp-bridge.el
+++ b/lsp-bridge.el
@@ -2066,11 +2066,14 @@ SymbolKind (defined in the LSP)."
   (evil-add-command-properties #'lsp-bridge-find-impl :jump t))
 
 (defun lsp-bridge--rename-file-advisor (orig-fun &optional arg &rest args)
-  (let ((current-file-name (expand-file-name (buffer-file-name)))
-        (new-name (expand-file-name (nth 0 args))))
+  (let* ((current-file-name (buffer-file-name))
+         (current-file-name (and current-file-name
+                                 (expand-file-name current-file-name)))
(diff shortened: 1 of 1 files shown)


lsp-bridge.el
```diff
@@ -2066,9 +2066,11 @@ SymbolKind (defined in the LSP)."
   (evil-add-command-properties #'lsp-bridge-find-impl :jump t))
 
 (defun lsp-bridge--rename-file-advisor (orig-fun &optional arg &rest args)
-  (when (and lsp-bridge-mode
-             (boundp 'acm-backend-lsp-filepath))
-    (let ((new-name (expand-file-name (nth 0 args))))
+  (let ((current-file-name (expand-file-name (buffer-file-name)))
```
manateelazycat, line 2069: If buffer-file-name is nil, expand-file-name will throw error `Wrong type argument: stringp, nil`

lsp-bridge.el
```diff
@@ -2066,9 +2066,11 @@ SymbolKind (defined in the LSP)."
...
+        (new-name (expand-file-name (nth 0 args))))
+    (when (and lsp-bridge-mode
+               (boundp 'acm-backend-lsp-filepath)
+               (file-equal-p current-file-name new-name))
```
manateelazycat, line 2073: Should use `string-equal`, because `file-euqal-p` will return `unspecified`, `unspecified` is True in Elisp.

manateelazycat (changes requested):

PR #692, opened 2023-07-28
improve copilot

kongds:
1. add node version check
2. add status check
3. improve npm package lookup


zerolfx:
https://emacs-china.org/t/lsp-bridge/20786/3276
顺便把这个也修了吧

PR #701, opened 2023-08-22
Support for async telega--searchChatMembers in the telega 0.8.162

zevlg:
Current implementation for username completion in telega have few issues:
1. It has  sync variant of fetching member list, because telega does not provide async API to fetch chat members
2. It uses potentially dangerous `buffer-list-update-hook` to do the job
3. It does some  job even if `lsp-bridge-mode` is disabled

This commit tries to resolve all these issues


manateelazycat:
great patch

PR #717, opened 2023-09-17
Add support for `clojure*-ts-mode`

motform:
This PR adds support for [`clojure-ts-mode`](https://github.com/clojure-emacs/clojure-ts-mode), the next generation of clojure modes.

I was not sure how to do the indentation since the line becomes very long, so I left it as simple as possible. Can change it to align by splitting the `ts`-modes to a new line if that is prefered.


manateelazycat:
Thanks for patch, great job. ;)

PR #722, opened 2023-10-03
provide diagnostic interface for flymake/flycheck frontend

liuyinz:
Integration with [flymake-bridge](https://github.com/liuyinz/flymake-bridge) to provide interface for traditional tools.

### DIFF
--- a/lsp-bridge-diagnostic.el
+++ b/lsp-bridge-diagnostic.el
@@ -90,6 +90,17 @@
   :type 'float
   :group 'lsp-bridge)
 
+(defcustom lsp-bridge-diagnostic-enable-overlays t
+  "If non-nil, rendering overlays in lsp-bridge-diagnostic byself.
+Set to nil if flymake/flycheck frontend is used instead."
+  :type 'boolean
+  :group 'lsp-bridge)
+
+(defcustom lsp-bridge-diagnostic-update-hook '()
+  "The hook for diagnostics updates."
+  :type 'hook
+  :group 'lsp-bridge)
+
 (defface lsp-bridge-diagnostics-error-face
   '((t (:underline (:style wave :color "Red1"))))
   "Face error diagnostic."
@@ -149,6 +160,8 @@ You can set this value with `(2 3 4) if you just need render error diagnostic."
   :type 'list
   :group 'lsp-bridge)
 
+(defvar-local lsp-bridge-diagnostic-records nil)
+
 (defvar-local lsp-bridge-diagnostic-overlays '())
 
 (defvar lsp-bridge-diagnostic-frame nil)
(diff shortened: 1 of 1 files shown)


manateelazycat:
First, thanks for patch. ;)

You can merge flymake-bridge and then use an option to control whether to render diagnostic with the lsp-bridge default mechanism or with flymake, so that Python only calls the --render function, not call two elisp function in this PR.

liuyinz:
Fine, I'll try. By the way, it there any methods to detect whether current buffer is connected to some lsp server and whether this server's diagnostic feature is enable?

manateelazycat:
We check lsp server capability at here https://github.com/manateelazycat/lsp-bridge/blob/c0b44de2361ea95cd422a7ab11553d436f350366/core/lspserver.py#L565 , but haven't diagnostic provider in init response.

We monitor response textDocument/publishDiagnostics message at here https://github.com/manateelazycat/lsp-bridge/blob/c0b44de2361ea95cd422a7ab11553d436f350366/core/lspserver.py#L557

I suggestion run flymake-bridge-setup code when first call lsp-bridge-diagnostic--render

liuyinz:
At first, I consider to merge flymake-bridge codes into lsp-bridge-diagnositc, however whether should flycheck become a optional dependency either in the future if somebody try to add the same thing?

I decide to write as  independent package:
```elisp
(setq lsp-bridge-diagnositc-use-builtin nil)  ;; disable default diagnostic overlays feature
(add-hook 'lsp-bridge-mode-hook #'flymake-bridge-setup) ;; enable flymake frontend
```
Flycheck could be implemented in same way and don't add extra dependency for lsp-bridge. What do you think @manateelazycat ?

manateelazycat:
Patch looks good， yes， i agree with you

lsp-bridge-diagnostic.el
```diff
@@ -169,13 +182,18 @@ You can set this value with `(2 3 4) if you just need render error diagnostic."
...
-                                      (diagnostic-number (length diagnostics)))
-                                  (dolist (diagnostic diagnostics)
-                                    (let ((severity (plist-get diagnostic :severity)))
-                                      (unless (member severity lsp-bridge-diagnostic-hide-severities)
+                                (setq-local lsp-bridge-diagnostic-records
+                                            (seq-remove
+                                             (lambda (diagnostic)
+                                               (member (plist-get diagnostic :severity) lsp-bridge-diagnostic-hide-severities))
+                                             diagnostics))
+                                (run-hooks 'lsp-bridge-diagnostic-update-hook)
```
manateelazycat, line 190: Those code only need when `lsp-bridge-diagnostic-enable-overlays` is nil.

manateelazycat (changes requested):

lsp-bridge-diagnostic.el
liuyinz: - I think keep update-hook unconditionally is a better idea to developers, add-hook would be far better than advice-add with -render function to use. 
- The var `lsp-bridge-diagnostic-records` is reused in  render overlays later to avoid filter by hide-severity twice, not totally redundancy.
-  I think save the originial diagnostics data everytime is quite important, so that developers could get much more infomations such as  error/wanring/info code: e.g. [TS2045]  to search for documentations, diagnositc overlays lose these informations. There is a max-number limit (50 by default) for diagnostics, would it slow-down the performance? I don't know, I guess rendering these overlays would take much more time than save values, maybe we need test.

manateelazycat: The reason why diagnostic information is limited to 50 is because when Emacs encounters thousands of diagnostic messages, drawing so many overlays, Emacs itself will freeze

liuyinz: I could add wrong code info to overlay message, would it be better?

liuyinz: @manateelazycat final version is pushed. Only add three lines for render function, have a look!

manateelazycat:
I have test patch in local, works fine, thank you. ;)

liuyinz:
Thanks for your briliant lsp-bridge too!

PR #738, opened 2023-10-12
fix void-function return

getong:
I enable `lsp-bridge-find-def-select-in-open-windows`, but error occurs:
```
ERROR:epc:ReturnError([Symbol('void-function'), Symbol('return')])
```
I test the code in my repo, and this pr works.


manateelazycat:
Thanks for patch.

PR #743, opened 2023-10-19
Fix :type and add :safe function to defcustom variables

zonuexe:
Add [`:safe` functions to defcustom](https://www.gnu.org/software/emacs/manual/html_node/elisp/Variable-Definitions.html#index-safe_002c-defcustom-keyword) to enable [local variables](https://www.gnu.org/software/emacs/manual/html_node/elisp/File-Local-Variables.html).

This isn't important for all custom variables, but it's useful to be able to override them on a per-project basis in [`.dir-locals.el`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Directory-Local-Variables.html).

### DIFF
--- a/lsp-bridge.el
+++ b/lsp-bridge.el
@@ -161,92 +161,112 @@ Setting this to nil or 0 will turn off the indicator."
     "python-black-buffer" "acm-complete-or-expand-yas-snippet" "acm-insert-number-or-complete-candiate"
     "yank" "string-rectangle" "query-replace" "grammatical-edit-unwrap")
   "If last command is match this option, stop popup completion ui."
-  :type 'cons
+  :type '(repeat string)
+  :safe #'listp
   :group 'lsp-bridge)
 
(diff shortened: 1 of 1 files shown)


lsp-bridge.el
```diff
@@ -161,92 +161,112 @@ Setting this to nil or 0 will turn off the indicator."
...
 (defcustom lsp-bridge-popup-documentation-buffer " *lsp-bridge-hover*"
   "Buffer for display hover information."
   :type 'string
+  :safe #'stringp
   :group 'lsp-bridge)
 
 (defcustom lsp-bridge-disable-backup t
   "Default disable backup feature, include `make-backup-files' `auto-save-default' and `create-lockfiles'."
   :type 'boolean
+  :safe #'stringp
```
manateelazycat, line 189: 这里为啥是stringp

zonuexe: Thank you, I fixed it.

manateelazycat:
感谢补丁， 我还不知道 :safe 关键字。 ;)

PR #750, opened 2023-10-24
Lsp bridge peek file content scroll margin

Moerliy:
Wanted to have a possibility to set the amount of lines that the peek window is moved up and down.


manateelazycat:
Thanks for patch.

PR #754, opened 2023-10-30
fix: choose lanugage id based on extension when using css lsp

braineo:
@manateelazycat  if `lsp-bridge/langserver/vscode-css-language-server.json` was in Python, then the server specific code can live in each config file. But here is the most reasonable place to add the fix atm.

### DIFF
--- a/core/lspserver.py
+++ b/core/lspserver.py
@@ -391,6 +391,16 @@ class LspServer:
         return uri
 
     def send_did_open_notification(self, fa: "FileAction"):
+        # vscode-css-language-server uses languageId to initialize from different entry points
+        if self.server_info["name"] == "vscode-css-language-server":
+            _, extension = os.path.splitext(fa.filepath)
+            if extension == ".less":
+                self.server_info["languageId"] = "less"
+            elif extension == ".scss" or extension == ".sass":
+                self.server_info["languageId"] = "scss"
+            else:
+                self.server_info["languageId"] = "css"
+
         self.sender.send_notification("textDocument/didOpen", {
             "textDocument": {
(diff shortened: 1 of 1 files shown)


core/lspserver.py
```diff
@@ -500,6 +511,7 @@ def handle_workspace_configuration_request(self, name, request_id, params):
             sessionSettings = settings.get(section, {})
 
             if self.server_info["name"] == "vscode-eslint-language-server":
+                sessionSettings = copy.copy(settings)
```
manateelazycat, line 514: why need copy

braineo: Not sure if somewhere else will modify workspace config, resulting in changing the server info secretly. It is just defensive — we can remove it if not needed.

manateelazycat:
I have push new commit https://github.com/manateelazycat/lsp-bridge/commit/086a363492151e9292ace54565275e74f5ff0ca3

In new commit, we check extension in languageIds first, avoid write lsp server name in lspserver.py

manateelazycat:
> > I have push new commit [086a363](https://github.com/manateelazycat/lsp-bridge/commit/086a363492151e9292ace54565275e74f5ff0ca3)我已经推送了新的提交 086a363
> > In new commit, we check extension in languageIds first, avoid write lsp server name in lspserver.py在新提交中，我们首先检查languageIds中的扩展，避免在 lspserver.py 中写入lsp服务器名称
> 
> Thank you. How about the commit fixing eslint server? I meant we can take out the copy.copy but the whole workspace setting needs to be filled.谢谢。提交修复 eslint 服务器怎么样？我的意思是我们可以取出 copy.copy，但需要填写整个工作区设置。

Please send new PR if have issue on eslint server. ;)

PR #756, opened 2023-10-31
fix: fix eslint lsp config that workspace config requires full settings

braineo:
eslint lsp now needs full session settings


manateelazycat:
Thanks.

PR #757, opened 2023-10-31
fix: support multiline diagnostics

braineo:
rust analyzer can return diagnostics with start and end in different lines. This PR fixes diagnostics list to show entries across multiple lines.


manateelazycat:
thanks man

Issue #764, opened 2023-11-04
Support opening documentation in a dedicated buffer


Zander671:
Hello,
I like to be able to open the documentation for a specific symbol in a separate window that I use split screen while editing. However, right now lsp-bridge seems to only support reporting documentation in a popup window. I was wondering if you guys would be willing to support adding another method of viewing documentation.
As a note, I am willing and ready to work on this if it is something you guys are interested in.

manateelazycat:
Sure, PR are welcome ;)

PR #766, opened 2023-11-05
Add dedicated persistent documentation buffer

Zander671:
Implement change proposed in #764


manateelazycat:
I have push commit https://github.com/manateelazycat/lsp-bridge/commit/838212a45772a4987d2d3fe1c8ece1fab78c79af

Change command name, use "hover" protocol.

We need add new key in all config under multiserver directory, but "Documentation" is not normal LSP protocol, it will confuse developer.

PR #780, opened 2023-11-14
Python3.12 does not ship with distutils. Use parse_version

triHims:


manateelazycat:
Can you write a patch to determine the version of Python? I'm worried that writing this will cause all versions of Python below 3.12 to not work

triHims:
updated the commit

manateelazycat:
Thanks

PR #793, opened 2023-11-23
Ctags

werhner:
1. support asynchronous ctags completion backend
2. make 'eval_in_emacs' from remote preserve order


manateelazycat:
Thanks for patch. ;)

PR #805, opened 2023-12-04
fix: filter diagnostics within the range to send to code action request

braineo:
without this fix, code action at point will list all the possible actions for the file regardless of the `range` argument in eslint server

this fix allows the server to filter the return value based on cursor position and selected range


manateelazycat:
Thanks for patch.

PR #820, opened 2024-01-02
reconnect remote host when socket is broken

werhner:
reconnect remote host when socket is broken


manateelazycat:
Thanks for patch.

PR #821, opened 2024-01-03
Support new features for finding refereneces

norris-young:
1. Delete other windows when finding references, default t.
2. Open file in the original window if we do not delete other windows, default nil.

### DIFF
--- a/lsp-bridge-ref.el
+++ b/lsp-bridge-ref.el
@@ -109,6 +109,20 @@ Default is enable, set this variable to nil if you don't like this feature."
   :type 'boolean
   :group 'lsp-bridge-ref)
 
+(defcustom lsp-bridge-ref-delete-other-windows t
+  "Delete other windows while finding references.
+
+Default is enable, set this variable to nil if you don't like this feature."
+  :type 'boolean
+  :group 'lsp-bridge-ref)
+
+(defcustom lsp-bridge-ref-open-file-in-request-window nil
+  "Open file in the window which requested to find references.
+
+Default is nil, set this variable to t if you like this feature."
+  :type 'boolean
+  :group 'lsp-bridge-ref)
+
 (defface lsp-bridge-ref-font-lock-header-line-text
   '((t (:foreground "Green3" :bold t)))
   "Face for header line text."
@@ -180,6 +194,9 @@ used to restore window configuration after finish search.")
 (defvar lsp-bridge-ref-buffer-point-before-search nil
   "Save buffer point before search,
 used to restore buffer point after finish search.")
+
+(defvar lsp-bridge-ref-request-search-window nil
+  "Save the window which requested to find reference")
 
 (defvar lsp-bridge-ref-window-configuration-before-apply nil
   "Save window configuration before apply changed,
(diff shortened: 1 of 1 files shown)


norris-young:
I don't want to delete other windows when finding references and still need files being opened in the original buffer.
Add 2 customizable variable to make it available and keep the default behavior unchanged.

lsp-bridge-ref.el
```diff
@@ -941,12 +967,24 @@ Function `move-to-column' can't handle mixed string of Chinese and English corre
...
   (when lsp-bridge-ref-window-configuration-before-search
-    (set-window-configuration lsp-bridge-ref-window-configuration-before-search)
+    ;; Restore window configuration before search if we delete-other-window before search.
+    ;; Or if we are inserting current line and the original window was deleted, we need to restore the window.
+    (if (or lsp-bridge-ref-delete-other-windows
+            (and insert
+                 (not (window-valid-p lsp-bridge-ref-requst-search-window))))
+        (set-window-configuration lsp-bridge-ref-window-configuration-before-search)
+      ;; Other we just delete the lsp-bridge-ref window, and select the original window.
+      (progn
```
manateelazycat, line 978: Don't need progn in elisp else branch.

lsp-bridge-ref.el
```diff
@@ -181,6 +195,9 @@ used to restore window configuration after finish search.")
   "Save buffer point before search,
 used to restore buffer point after finish search.")
 
+(defvar lsp-bridge-ref-requst-search-window nil
```
manateelazycat, line 198: Should ```lsp-bridge-ref-request-search-window``` ? I found miss `e` in `request`.

lsp-bridge-ref.el
```diff
@@ -941,12 +967,24 @@ Function `move-to-column' can't handle mixed string of Chinese and English corre
...
+        (delete-window)
+        (select-window lsp-bridge-ref-requst-search-window)))
     (goto-char lsp-bridge-ref-buffer-point-before-search)
     (setq lsp-bridge-ref-window-configuration-before-search nil)
-    (setq lsp-bridge-ref-buffer-point-before-search nil)))
+    (setq lsp-bridge-ref-buffer-point-before-search nil))
+  ;; clear saved window which requested search.
+  (if lsp-bridge-ref-open-file-in-requset-window
+      (setq lsp-bridge-ref-requst-search-window nil))
+  )
```
manateelazycat, line 987: If i understand correctly, lsp-bridge-ref-request-search-window should reset to nil always, don't need ```if``` check here.

lsp-bridge-ref.el
```diff
@@ -109,6 +109,20 @@ Default is enable, set this variable to nil if you don't like this feature."
...
   :group 'lsp-bridge-ref)
 
+(defcustom lsp-bridge-ref-delete-other-windows t
+  "Delete other windows while finding references.
+
+Default is enable, set this variable to nil if you don't like this feature."
+  :type 'boolean
+  :group 'lsp-bridge-ref)
+
+(defcustom lsp-bridge-ref-open-file-in-requset-window nil
```
manateelazycat, line 119: Should ```lsp-bridge-ref-open-file-in-request-window``` ?

manateelazycat (changes requested):

lsp-bridge-ref.el
norris-young: fixed

lsp-bridge-ref.el
norris-young: Fixed

lsp-bridge-ref.el
norris-young: Yeah, you are right.

lsp-bridge-ref.el
norris-young: Fixed

norris-young:
Besides, I changed the argument name of `lsp-bridge-ref-quit` from `insert` to `restore`, which should be more suitable.

manateelazycat:
Thanks for patch.

PR #829, opened 2024-01-12
Fix missing import completions for same symbol name in multiple files.

runekaagaard:
Fixes @826


manateelazycat:
Thanks for help . ;)

PR #830, opened 2024-01-13
fix lsp-bridge-try-completion bug with lsp-bridge-enable-log set to t

yatsky:
This bug causes `wrong-number-of-arguments` error when `lsp-bridge-enable-log` is set to `t` and will cause Emacs remove `lsp-bridge-monitor-post-command` from `post-command-hook`.


manateelazycat:
thanks

PR #842, opened 2024-01-28
Fix remote server connection

norris-young:
Add a `heartbeat` mechanism ffd0b982be317ac2c16b1ac22808c092414fe4ee for keeping remote connection alive to resovle #831 .
In the process of analyzing #831 ,I found some other bugs to fix here:

1. 431171c7bf192e7b809a4d970b65cde2d68d8609 `lsp-bridge-is-remote-file` needs synchronization with the server, so it is not real-time. We should not send requests for these remote files in updating to local servers or we will get unexpected errors. Before remote sync done, we just skip requests for these files.
2. c328cc8fc4627d72672c02ba4b88e713000009b1 We should call `sync_tramp_remote` with pure text buffer name.
3. ceb2f189a461c8c2efa0a2c43240d83d2f4ad408 When we have both opened remote files and local files and do `lsp-bridge-restart-process` in a local file buffer, we would get errors if we do not remove remote files on indexing.

Besides, the code for LspBridge remote server seems to be in a mess and hard to understand. So I tried to do a refactor. 
6ce0a1a773826af8fa7bcaa297799fed6a653be3 99476ec9a0d328132f831c244dbdc8728c912f68 b1200f0882d55822d044071c98dae6c33984cf94

### DIFF
--- a/core/remote_file.py
+++ b/core/remote_file.py
@@ -25,6 +25,7 @@ import glob
 import json
 import socket
 import traceback
+import time
 from core.utils import *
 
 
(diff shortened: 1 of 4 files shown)


manateelazycat:
thanks for patch

manateelazycat:
我要紧急撤回这一批补丁， 这一批补丁升级以后， 导致 Python 语言无法补全。

请把您要提交的补丁， 按照每一个功能， 单独提交， 这样更容易测试， 也更容易合并。

manateelazycat:
我已经提交了 https://github.com/manateelazycat/lsp-bridge/commit/5baf390b5fb2a7d821ba73cf401c6c8ec419cb4c ， 撤销了这个PR的所有代码。

请把功能单开拆分提交， 一个功能一个补丁， 并保障远程补全可以工作的情况下， 本地代码补全也要测试。

norris-young:
好的 大概已经找到问题了 我再测试测试 有空再分开提过来

norris-young:
已经尽量拆分开了
没有依赖关系的patch我先推了 #847 #848 #849 #850 
有依赖关系的后面一个一个推过来

manateelazycat:
好的，我这几天堵在高速上，等我到酒店有空了，我来看这些补丁，辛苦了，辛苦了

PR #847, opened 2024-02-02
Skip calling file api when we are updating tramp file info

norris-young:
When a tramp file is opened, its information is updated by calling `lsp-bridge-sync-tramp-remote` and will be finished in `lsp-bridge-update-tramp-file-info`. Before it is done, `lsp-bridge-remote-file-flag` is nil and related remote file info variables are not set. If we call file api this time, it will send the request to local lsp server(local epc call). Because the file is a remote file in fact, we won't get right result frome local server and it might leads to some unpredicatable exception. So we add check in `lsp-bridge-call-file-api` to skip the call if update is not finished.


manateelazycat:
Local test pass, merged it, thanks.

PR #848, opened 2024-02-02
Get only text string for tramp file name

norris-young:
`buffer-file-name` might have properties and we should only pass text string without any properties. Or we will get an error fo wrong number of arguments.


manateelazycat:
Local test pass.

PR #849, opened 2024-02-02
Remove remote files on indexing local files

norris-young:
In the branch for indexing local files, it didn't remove local files from `buffer-list`. Calling `search_file_words_index_files` for local with remote file names will bring error. So we need to add a check rule `(file-remote-p elt)`.


manateelazycat:
Local test pass

PR #858, opened 2024-02-12
Unify receive_remote_message for different file servers

norris-young:
Add a queue for elisp channel to unify remote message process and parse json in common code instead of each handler.


norris-young:
This PR has dependency which needs to be merged before it.

PR #862, opened 2024-02-13
Support heartbeat for remote connections

norris-young:
If client does not send messages to the server for more than a specific time(e.g. 2h12min for me), the server will close the socket while the client can still send messages without error. However, the server will not receive and handle any more messages. In order to solve this issue, we add a mechanism to send heartbeats to the server to keep the connection alive.

### DIFF
--- a/core/remote_file.py
+++ b/core/remote_file.py
@@ -25,6 +25,7 @@ import glob
 import json
 import socket
 import traceback
+import time
 from core.utils import *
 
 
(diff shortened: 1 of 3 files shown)


norris-young:
Besides, add more debug info for remote usage.

manateelazycat:
已经合并了， 可以把心跳的变量写到两个README文件中， 说不定其他人也需要呢。

PR #863, opened 2024-02-13
Support jumping to references for remote files without tramp

norris-young:
1. Tweak the keys of the dict `host_names` storing ssh connection info so it will be compatible with class `paramiko.config.SSHConfigDict` which inherits the internal dict.
2. Also update remote information when `open_remote_file` in order to support jumping to references without using tramp.


manateelazycat:
感谢补丁， 已经合并

PR #871, opened 2024-02-24
Do not sync remote for local files with sudo

norris-young:
Tramp files with 'sudo' like methods are local files in fact. We should not try to sync tramp remote for these files.


norris-young:
刚刚发现好像把localhost name加入到`lsp-bridge-tramp-blacklist`里也可以解决问题，猫大看看是否还要这个patch，好像不是特别必要了

manateelazycat:
感谢补丁，已经合并

PR #873, opened 2024-02-25
Org roam backend

pastnontra:
Add new `acm-backend-org-roam`.

- New file: `acm-backend-org-roam.el`.
- New `input-prefix-bound` type: `org-roam`, it allows to complete non-ASCII string containing space in org roam brackets.
- Change some logic during inputting and expanding.

Current completion is still imperfect, it won't pop up completion after I type space; it also seems I need to have exact matched string after I press space.
E.g. If I have a node named "foo bar", "fo ba|" won't show the candidate; it's still case insensitive, "FOo ba|" shows that candidate.

Not use Python backend to speedup. Org roam is complicated, I'm not sure if I'll break up something, since the performance (with few hundreds nodes) is acceptable for me, maybe one day.

Anyway, it's usable enough for me.


manateelazycat:
Thanks for patch, more PR are welcome. ;)

PR #879, opened 2024-03-04
fix: always set diagnostics for first paint of flycheck/flymake update

braineo:
since flycheck / flymake can be enabled later with some interactive functions depending on the use cases.

with the `(unless lsp-bridge-diagnostic-enable-overlays)` check, the diagnostics are not available until the users start typing anything to trigger some lsp command.

if there wasn't a good reason (memory? maybe?), can we always enable it?


manateelazycat:
It's okay to record it.

Issue #877, opened 2024-03-03
Oderless causes (wrong-type-argument stringp (seq (group 100)))


pastnontra:
After I updated `orderless`, I can't complete `yasnippet` candidates if I `(setq acm-candidate-match-function 'orderless-flex)`.

It could be introduced by this commit https://github.com/oantolin/orderless/commit/674a1c75d5861b36de29b2fa1957146a3ee34163.

- Emacs version: GNU Emacs 29.1 (build 1, x86_64-apple-darwin22.6.0, Carbon Version 169 AppKit 2299.7)
- Lsp-bridge version: master

Here is the backtrace when I typing in `emacs-lisp-mode`:
```lisp
Debugger entered--Lisp error: (wrong-type-argument stringp (seq (group 100)))
  string-match((seq (group 100)) "(add-hook 'major-mode-hook ...)" nil t)
  string-match-p((seq (group 100)) "(add-hook 'major-mode-hook ...)")
  acm-candidate-fuzzy-search("d" "(add-hook 'major-mode-hook ...)")
  (closure ((keyword . "d")) (s) (acm-candidate-fuzzy-search keyword (car (acm-backend-yas--template-cons s))))(#s(yas--template :key "hook" :content "(defun ${1:major-mode}-hook-setup ()\n  $0\n  (local..." :name "(add-hook 'major-mode-hook ...)" :condition nil :expand-env nil :load-file "/Users/pastnontra/.emacs.d/snippets/emacs-lisp-mod..." :save-file nil :keybinding nil :uuid "(add-hook 'major-mode-hook ...)" :menu-binding-pair ((menu-item "(add-hook 'major-mode-hook ...)" (lambda nil (interactive) (yas--expand-or-visit-from-menu 'emacs-lisp-mode "(add-hook 'major-mode-hook ...)")) :keys "hook⇥")) :group nil :perm-group nil :table #s(yas--table :name "emacs-lisp-mode" :hash #<hash-table equal 21/65 0x1fe4c2786689> :uuidhash #<hash-table equal 29/65 0x1fe4c278695f> :parents nil :direct-keymap (keymap))))
[...]

PR #880, opened 2024-03-04
Let acm-candidate-fuzzy-search accept rx regexp.

pastnontra:
fix issue #877

### DIFF
--- a/acm/acm.el
+++ b/acm/acm.el
@@ -397,8 +397,10 @@ So we use `minor-mode-overriding-map-alist' to override key, make sure all keys 
 
 (defun acm-candidate-fuzzy-search (keyword candidate)
   "Fuzzy search candidate."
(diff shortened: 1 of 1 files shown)


acm/acm.el
```diff
@@ -397,8 +397,10 @@ So we use `minor-mode-overriding-map-alist' to override key, make sure all keys
 
 (defun acm-candidate-fuzzy-search (keyword candidate)
   "Fuzzy search candidate."
-  (string-match-p (funcall acm-candidate-match-function (downcase keyword))
-                  (downcase candidate)))
+  (let ((result (funcall acm-candidate-match-function (downcase keyword))))
```
pastnontra, line 400: Maybe name this variable "regexp", so we don't confuse with completion results?

manateelazycat:
Thanks for patch, good job. ;)

PR #885, opened 2024-03-08
Fix completion error "Wrong type argument: integer-or-marker-p, nil"

zhoupen9:
Remove unused `goto-char yas-snippet-end` in `acm-backend-lsp.el` which cause completion failure.


manateelazycat:
Thanks for patch, I have fixed it in commit https://github.com/manateelazycat/lsp-bridge/commit/935f7181102182045ba7d203d22a4d717bcd127d

PR #889, opened 2024-03-15
Add ruby-ts-mode support

yonta:
Added support for ruby-ts-mode.
ruby-ts-mode has become built-in packages in Emacs 29 and later.


acm/acm-backend-codeium.el
getong, line 1: line 121 is already here, duplicated

yonta: Ooops!
Thanks to delete duplicated line :)

https://github.com/manateelazycat/lsp-bridge/pull/891

PR #890, opened 2024-03-15
feat: add emacs-jupyter completion backend for acm in org babel

sawyerzheng:
# add a emacs-jupyter backend for acm babel completion

1. This backend can be activated by set `acm-enable-jupyter` to `t`
2. This backend only work when in a org-mode source code block and its language starts with `"jupyter"`
3. It still can make completion even though the `import` statement in a previous source code block
4. It need to start the jupyter in the org mode buffer first

image

### DIFF
--- a/acm/acm-backend-jupyter.el
+++ b/acm/acm-backend-jupyter.el
@@ -22,9 +22,27 @@
     keyword
     acm-backend-jupyter-items)))
 
-(defun lsp-bridge-jupyter-record (candidates)
-  (setq-local acm-backend-jupyter-items candidates)
-  (lsp-bridge-try-completion))
+(defun acm-backend-jupyter-record (current-symbol)
+  (interactive)
+  (unless (fboundp 'jupyter-org-completion-at-point)
+    (require 'jupyter)
+    (require 'jupyter-org-client)
+    (require 'ob-jupyter))
+  (setq-local acm-backend-jupyter-items nil)
+  (let* ((candidates (mapcar #'substring-no-properties (all-completions "" (nth 2 (jupyter-org-completion-at-point)))))
+         (candidates (if (stringp current-symbol)
+                         (seq-filter (apply-partially #'string-prefix-p current-symbol) candidates)
+                       candidates))
+         (candidates-with-metadata (mapcar (lambda (candidate)
+                                             (list :key candidate
+                                                   :icon ""
+                                                   :label candidate
+                                                   :displayLabel candidate
+                                                   :annotation "jupyter"
+                                                   :backend "jupyter"))
+                                           candidates)))
+    (when candidates
+      (lsp-bridge-search-backend--record-items "jupyter" candidates-with-metadata))))
 
 (defun acm-backend-jupyter-clean ()
   (setq-local acm-backend-jupyter-items nil)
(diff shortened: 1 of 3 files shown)


acm/acm-backend-jupyter.el
```diff
@@ -0,0 +1,34 @@
...
+(defun acm-backend-jupyter-candidates (keyword)
+  (acm-with-cache-candidates
+   acm-backend-jupyter-cache-candiates
+   (acm-candidate-sort-by-prefix
+    keyword
+    acm-backend-jupyter-items)))
+
+(defun lsp-bridge-jupyter-record (candidates)
+  (setq-local acm-backend-jupyter-items candidates)
+  (lsp-bridge-try-completion))
```
manateelazycat, line 27: I don't see where the function lsp-bridge-jupyter-record is called

lsp-bridge.el
```diff
@@ -2531,6 +2541,27 @@ We need exclude `markdown-code-fontification:*' buffer in `lsp-bridge-monitor-be
...
+                                               (list :key text
+                                                     :icon ""
+                                                     :label text
+                                                     :displayLabel text
+                                                     :annotation "jupyter"
+                                                     :backend "jupyter")))
+                                           candidates)))
+    (when candidates
+      (lsp-bridge-search-backend--record-items "jupyter" candidates-with-metadata)
+      )))
```
manateelazycat, line 2563: Those code should move to acm-backend-jupyter and rename with function name acm-backend-jupyter-record

sawyerzheng: If I rename these codes to function acm-backend-jupyter-record in acm-backend-jupyter.el file, how can I call it in `lsp-bridge-complete-other-backends` function?
Sorry, I just do not know how to trigger the `acm-backend-jupyter-record` function. It seems it is called by `.py` files for other backend.

If I call it with the following code, the pop menu candidate will not update as I typing：

``` elisp
    (when (and acm-enable-jupyter
               (lsp-bridge-process-live-p)
               (and lsp-bridge-enable-org-babel
                    (eq major-mode 'org-mode)
                    (not (lsp-bridge-is-org-temp-buffer-p)))
               (org-in-src-block-p 'inside)
               (string-prefix-p "jupyter" (plist-get (car (cdr (org-element-context))) :language)))
      ;; (lsp-bridge-call-async "jupyter" current-word)
      (acm-backend-jupyter-record nil)
      )
```

```elisp
(defun acm-backend-jupyter-record (candidates)
  (unless (fboundp 'jupyter-org-completion-at-point)
    (require 'jupyter)
    (require 'jupyter-org-client)
    (require 'ob-jupyter))
  (setq-local acm-backend-jupyter-items nil)
  (let* ((candidates (all-completions "" (nth 2 (jupyter-org-completion-at-point))))
         (candidates-with-metadata (mapcar (lambda (candidate)
                                             (let ((text (substring-no-properties candidate)))
                                               (list :key text
                                                     :icon ""
                                                     :label text
                                                     :displayLabel text
                                                     :annotation "jupyter"
                                                     :backend "jupyter")))
                                           candidates)))
    (when candidates
      (lsp-bridge-search-backend--record-items "jupyter" candidates-with-metadata))))
```

sawyerzheng:
new commit, 
(1) moved lsp-bridge-jupyter-complete function to acm-backend-jupyter.el
(2) filtered the candidate with the `current-symbol` before cursor when current-symbol is a string 

image

manateelazycat:
Thanks

PR #912, opened 2024-04-16
Sync tramp info

werhner:
1. Accelerate the tramp file opening speed on already connected remote machines.
2. Adapt the remove_prefix function for Python versions prior to 3.9.
3. Use "ps | grep " to determine if the lsp_bridge.py process has started to prevent misjudgment.


manateelazycat:
Thanks for patch

PR #917, opened 2024-04-26
Add support for "tempel-trigger-prefix"

rennsax:
The variable `tempel-trigger-prefix` controls which prefix need to be input first before the completion can recognize it. For example, if there is a snippet "header" returned by `tempel--templates`, with `tempel-trigger-prefix` set to "t", "theader" will then expand to the snippet body.

This variable should be respected by ACM too.

The effects:

With prefix set, no tempel completion options show up:

image

If the prefix is entered, tempel snippets can be matched:

image

TAB, the completion expands:

image


manateelazycat:
thanks

PR #920, opened 2024-04-26
support code action and code format on tramp file

werhner:
support code action and code format on tramp file.


manateelazycat:
thanks，great work

PR #947, opened 2024-05-15
extra path sanitation

aetherspritee:
dont know if thats wanted but i apparently had a random list in one of my paths which tripped lsp-bridge up.


manateelazycat:
thanks

PR #956, opened 2024-05-29
Fix naming issues with copilot and codeium on Windows

CnsMaple:
Can't use copilot and codeium in Windows because of the name issue.


manateelazycat:
Thanks for patch

PR #957, opened 2024-05-31
Fix insert_spaces type in codeium.py

CnsMaple:
Sometimes `insert_spaces` is passed a list type parameter, which then causes codeium to crash. It needs to be of type bool.

Clip_2024-06-01_03-19-43


manateelazycat:
thanks

PR #958, opened 2024-06-01
set vue hybrideMode false

getong:
see https://github.com/joaotavora/eglot/discussions/1395#discussioncomment-9280456

设置vue hybrideMode 为false，这样volar才正常工作。


manateelazycat:
thanks

PR #960, opened 2024-06-05
Add nixd, new lsp server for Nix

arthsmn:
I'm not setting it as default, but I think it can be discussed, as the default one is abandoned and this one is much more powerfull than nil.

Edit: found out that rnix-lsp is not available in nixos-unstable or nixos-24.05 (latest stable), so it is basically not supported in NixOS anymore.


manateelazycat:
Thanks

PR #961, opened 2024-06-09
Don't hide completion menu when the exact match candidate is Org Roam

pastnontra:
Adjust relevant comment.
Fix indentations about Org roam.


manateelazycat:
请把缩进的部分去掉吧， 影响看补丁

PR #965, opened 2024-06-10
Add pyenv instructions to README

tjtrabue:
Python distributions installed via `pyenv` require special treatment to work with lsp-bridge. `pyenv` is primarily a shell utility, and does not always integrate well with other processes since it assumes that the user has sourced extra definitions into their shell process before using it, and places a number of shim files
(which are just shell scripts that adjust the actual command to run based on the selected `pyenv` Python SDK version) before their executable counterparts in the PATH hierarchy. This messes with lsp-bridge, which requires an actual `python3` executable to launch the LSP server process, not a shim.

To fix this issue, users must set their `lsp-bridge-python-command` variable to the actual path to the `python3` executable for their selected Python version. This patch provides instructions on how to do that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,25 @@ and add this to your `config.el`
 
 and run `doom sync` to install it.
 
+### `pyenv` configuration
+
+If you use a Python distribution installed via `pyenv`, you must adjust your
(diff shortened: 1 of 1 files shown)


manateelazycat:
Thank you

PR #968, opened 2024-06-12
Container: handle visitied file modtime & Update README

nohzafk:
Editng container file:
- Handle visitied file modified time, so emacs won't warn about file has been changed since visited or saved.
- Support auto refresh file content when formatter format the file

Update README 
- how to enable lsp-bridge devcontainer
- how to enable formatting container file
- how to add a reminder that currently editing a remote file


manateelazycat:
Thanks for patch.

PR #980, opened 2024-06-27
Use `bound-and-true-p` instead of `boundp`

zonuexe:
The code can be made much simpler by using the `bound-and-true-p` macro [defined in `bindings.el`](https://github.com/emacs-mirror/emacs/blob/2078ce4220ba153a685962e7480f3ac1fb2e0b36/lisp/bindings.el#L765-L771).
Despite its name, it returns the value bound to the variable, not a boolean (`t` or `nil`).

### DIFF
--- a/lsp-bridge.el
+++ b/lsp-bridge.el
@@ -828,8 +828,7 @@ So we build this macro to restore postion after code format."
      (forward-char (max (- current-column indent-column) 0))))
 
 (defun lsp-bridge-is-remote-file ()
-  (and (boundp 'lsp-bridge-remote-file-flag)
-       lsp-bridge-remote-file-flag))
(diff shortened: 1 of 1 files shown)


lsp-bridge.el
```diff
@@ -1491,7 +1483,7 @@ So we build this macro to restore postion after code format."
    ))
 
 (defun lsp-bridge--not-in-org-table ()
-  (not (and (boundp 'org-at-table-p)
+  (not (and (fboundp 'org-at-table-p)
```
zonuexe, line 1486: Since `org-at-table-p` is a function, not a variable, this code did not work as expected.

manateelazycat:
@zonuexe This patch cause CI failed, looks not good

PR #981, opened 2024-06-27
Update README.md

gchape:
It is now typescript-language-server rather than a typescript link. Since lsp for JavaScript is out of date and should be replaced with typescript, typescript and typescript-language-server are necessary.

React was added.

Removed EsLint, a feature that is not required and can be enabled optionally. The download path for ESLint was wrong.


manateelazycat:
Thanks

PR #993, opened 2024-07-20
feat: support treesit ts mode for csharp

kira-bruneau:
Adds support for the csharp tree-sitter mode included with emacs.

Also removes references to `csharp-tree-sitter-mode`, which is part of https://github.com/emacs-csharp/csharp-mode. It's been merged into emacs core and is no longer maintained as a separate repo.


manateelazycat:
thanks

PR #994, opened 2024-07-20
feat: support treesit ts mode for dockerfile

kira-bruneau:
Adds support for the docker tree-sitter mode included with emacs.


manateelazycat:
thanks

PR #995, opened 2024-07-20
feat: support treesit ts mode for nix

kira-bruneau:
Adds support for the nix tree-sitter mode: https://github.com/nix-community/nix-ts-mode/


manateelazycat:
thanks

PR #1003, opened 2024-07-29
Fixup getting project path on Windows

md11235:
Sometimes ‘git rev-parse —show-toplevel’ is used in core/utils.py to get project path for an existing source file. That git subcommand would convert a Windows style path like “f:/foo/bar.py” to “/f/foo” and cause unexpected behavior on Windows OS’s.

I’ve made a possible fix to that git behavior. Thank you for reviewing this patch.

### DIFF
--- a/core/utils.py
+++ b/core/utils.py
@@ -362,7 +362,13 @@ def get_project_path(filepath):
         import os
         dir_path = os.path.dirname(filepath)
         if get_command_result("git rev-parse --is-inside-work-tree", dir_path) == "true":
-            return get_command_result("git rev-parse --show-toplevel", dir_path)
+            path_from_git = get_command_result("git rev-parse --show-toplevel", dir_path)
+            if get_os_name() == "windows":
+                path_parts = path_from_git.split("/")
+                windows_path = path_parts[1] + ":/" + "/".join(path_parts[2:])
+                return windows_path
+            else:
+                return path_from_git
         else:
             return filepath
 


manateelazycat:
@md11235 
Thanks for patch, can you tell me what's value of `path_from_git` in your system?  And what's your target string for windows_path?

I think it's dangerous if we direct access `1` or `2` index in path_parts, will maybe crash lsp-bridge.

md11235:
Thank you for you reply. You're correct about this. 

Previously I only tested the git command provided MSYS2 (https://www.msys2.org/). And I've instead tested the git command provided by "Git for Windows" (https://gitforwindows.org/) just now and it behaved differently.

Suppose f:/foo is a local git repository. In MSYS2, `git rev-parse --show-toplevel f:/foo/bar.py` produces:

`/f/foo`

Yet in "Git For Windows", `git rev-parse --show-toplevel f:/foo/bar.py` produces:

`F:/foo`

So it's just the git command from MSYS2 when using the subcommand `rev-parse` has the problem of converting Windows-style path to a Unix-style one.

md11235:
Thank you for you reply. You're correct about this. 

Previously I only tested the git command provided MSYS2 (https://www.msys2.org/). And I've instead tested the git command provided by "Git for Windows" (https://gitforwindows.org/) just now and it behaved differently.

Suppose f:/foo is a local git repository. In MSYS2, `git rev-parse --show-toplevel f:/foo/bar.py` produces:

`/f/foo`

Yet in "Git For Windows", `git rev-parse --show-toplevel f:/foo/bar.py` produces:

`F:/foo`

So it's just the git command from MSYS2 when using the subcommand `rev-parse` has the problem of converting Windows-style path to a Unix-style one.

md11235:
Thank you for you reply. You're correct about this. 

Previously I only tested the git command provided MSYS2 (https://www.msys2.org/). And I've instead tested the git command provided by "Git for Windows" (https://gitforwindows.org/) just now and it behaved differently.

Suppose f:/foo is a local git repository. In MSYS2, using git version 2.45.2, `git rev-parse --show-toplevel f:/foo/bar.py` produces:

`/f/foo`

Yet in "Git For Windows", using git version 2.46.0.windows.1, `git rev-parse --show-toplevel f:/foo/bar.py` produces:

`F:/foo`

So it's just the git command from MSYS2 when using the subcommand `rev-parse` has the problem of converting Windows-style path to a Unix-style one.

md11235:
> Thank you for you reply. You're correct about this.
> 
> Previously I only tested the git command provided MSYS2 (https://www.msys2.org/). And I've instead tested the git command provided by "Git for Windows" (https://gitforwindows.org/) just now and it behaved differently.
> 
> Suppose f:/foo is a local git repository. In MSYS2, `git rev-parse --show-toplevel f:/foo/bar.py` produces:
> 
> `/f/foo`
> 
> Yet in "Git For Windows", `git rev-parse --show-toplevel f:/foo/bar.py` produces:
> 
> `F:/foo`
> 
> So it's just the git command from MSYS2 when using the subcommand `rev-parse` has the problem of converting Windows-style path to a Unix-style one.

Both these git commands were executed on Windows 10(Version 10.0.19045.4651).

core/utils.py
```diff
@@ -362,7 +362,13 @@ def get_project_path(filepath):
         import os
         dir_path = os.path.dirname(filepath)
         if get_command_result("git rev-parse --is-inside-work-tree", dir_path) == "true":
-            return get_command_result("git rev-parse --show-toplevel", dir_path)
+            path_from_git = get_command_result("git rev-parse --show-toplevel", dir_path)
+            if get_os_name() == "windows":
```
manateelazycat, line 366: This line should change to:

if get_os_name() == "windows" and path_from_git.startswith("/"):

manateelazycat (changes requested):

core/utils.py
manateelazycat: Or only run convert code when env is MSYS2

manateelazycat:
no， linux path always starts with /，you need check system is windows and path starts with /

linux path do not need convert code

core/utils.py
```diff
@@ -362,7 +362,17 @@ def get_project_path(filepath):
...
+                path_parts = path_from_git.split("/")
+                # if this is a Unix-style absolute path, which should be a Windows-style one
```
manateelazycat, line 368: no right

PR #1004, opened 2024-07-30
Remove set-visited-file-name for remote docker file

nohzafk:
Remove it will make things just work, keeping buffer as a non-file buffer
- aphelia formatter will work with remote file
- use `set-visited-file-name` will result in saving the file twice, and we have to deal with buffer `modtime`, and unsaved change.
- keep the same logic as remote SSH file

Update README about how to use formatter with remote file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,13 +176,13 @@ If you use `apheleia` as formatter, `lsp-bridge` now support auto formatting fil
 ```elsip
 (use-package! apheleia
   :config
-  (setq +format-with-lsp nil)
   ;; which formatter to use
   (setf (alist-get 'python-mode apheleia-mode-alist) 'ruff)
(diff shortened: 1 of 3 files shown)


manateelazycat:
Thanks for patch

manateelazycat:
@nohzafk Looks strange, https://github.com/manateelazycat/lsp-bridge/commit/d54ffdf127e92c0faae96e005026833c44007aa8 will make github action failed.

nohzafk:
is it Ok right now? the code has nothing special in it

manateelazycat:
It's not bug of lsp-birdge, it's issue https://github.com/purcell/setup-emacs/issues/49#issuecomment-2259930061

nohzafk:
thanks for the info!

PR #1007, opened 2024-07-31
should wait for signal before set it

nohzafk:
1. We should wait for the signal before setting it.

2. A race condition exists where init_search_backends may finish running before waiting for `init_search_backends_complete_event` to be set. If this happens, the waiting process will never finish, blocking the entire process.


manateelazycat:
Thanks

PR #1009, opened 2024-08-03
Support lsp-bridge in container correctly

nohzafk:
# Unify the logic for remote SSH file and docker file

When using `find-file` to open a file in container, the TRAMP buffer is now setup as the **lsp-bridge buffer**

- we treat the TRAMP buffer as a non-file buffer by setting `buffer-file-name` to nil
- previous impl is not correct (create a new buffer for docker file and close the corresponding TRAMP buffer)
 
# Support apheleia formatter in container
Update README for how to setup PATH for remote command execution, otherwise the formatter binary won't be found (for example `ruff`, even it is installed in container)

# Support find defition in container 

Support using `lsp-bridge-find-def` `lsp-bridge-find-def-return`  `lsp-bridge-find-def-other-window`  in container.

Text Matrix

| lsp-bridge-find-def-select-in-open-windows | docker file | lsp-bridge-find-def/-return                                                                | lsp-bridge-find-def-other-window |
| ------------------------------------------ | ----------- | ------------------------------------------------------------------------------------------ | -------------------------------- |
| nil                                        | yes         | pass - jump to another file and jump back<br>pass - jump to current file and jump back<br> | pass - open a new window         |
| nil                                        | no          | pass - jump to another file and jump back<br>pass - jump to current file and jump back<br> | pass - open a new window         |
| t                                          | yes         | jump to opened window - pass                                                               | pass - jump to opened window     |
| t                                          | no          | jump to opened window - pass                                                               | pass - jump to opened window     |

## issue
Under current implementation if jump to other window,` lsp-bridge-find-def-return` can't jump back to original window. It would be nice if we support jump back to original window.

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,9 +172,13 @@ start the devcontainer and use `file-find` `/docker:user@container:/path/to/file
 more detail please refer to [devcontainer-feature-emacs-lsp-bridge](https://github.com/nohzafk/devcontainer-feature-emacs-lsp-bridge).
 
 If you use `apheleia` as formatter, `lsp-bridge` now support auto formatting file on devcontainer.
+```elisp
+;; setup PATH for remote command execution
+(with-eval-after-load 'tramp
+  (add-to-list 'tramp-remote-path "~/.nix-profile/bin")
+  (add-to-list 'tramp-remote-path 'tramp-own-remote-path))
 
-```elsip
-(use-package! apheleia
+(use-package apheleia
   :config
   ;; which formatter to use
   (setf (alist-get 'python-mode apheleia-mode-alist) 'ruff)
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -172,9 +172,13 @@ start the devcontainer and use `file-find` `/docker:user@container:/path/to/file
 more detail please refer to [devcontainer-feature-emacs-lsp-bridge](https://github.com/nohzafk/devcontainer-feature-emacs-lsp-bridge).
 
 If you use `apheleia` as formatter, `lsp-bridge` now support auto formatting file on devcontainer.
+```elisp
+;; setup PATH for remote command execution
+(with-eval-after-load 'tramp
+  (add-to-list 'tramp-remote-path "~/.nix-profile/bin")
```
manateelazycat, line 178: lsp-bridge user must be install Nix?

nohzafk: container support is done by using the devcontainer feature which use nix to install lsp-bridge and language server.

I'm planing to create a separate repo for instructions and setup of how to use lsp-bridge in container. Will update README to point to the new repo later.

PR #1055, opened 2024-10-03
fix when dir-local-find-file return string first time

fuyu0425:
When searching .dir-locals.el first time in a project (not in dir-locals-directory-cache), dir-local-find-file will return string instead of a list


manateelazycat:
thanks

PR #1059, opened 2024-10-04
use os.path.normpath to remove the trailing / of project path

fuyu0425:
dir-locals-find-file will return project directory ending with /, which make os.path.basename to return empty string, so I add os.path.normpath to remove the trailing /.


manateelazycat:
Thanks for the patch

The correctness of the project_path path is too important for lsp-bridge. If dir-locals-find-file returns an incorrect value, I think it should be filtered on the elisp side and should not be processed on the python side.

fuyu0425:
ok, let me do that

PR #1061, opened 2024-10-05
fix texlab langagueIds

fuyu0425:
based on
https://github.com/latex-lsp/texlab/blob/master/crates/distro/src/language.rs


manateelazycat:
thanks

PR #1065, opened 2024-10-07
make lsp-bridge-diagnostic-hide-severities to works for list

fuyu0425:
Pass to python side and filter on python side.
This enables something like this to dynamically suppress some diagnostic.

```emacs-lisp
(defun leo/lsp-bridge-diagnostic-list-all ()
  (interactive)
  (let ((lsp-bridge-diagnostic-hide-severities '()))
    (lsp-bridge-diagnostic-list)))
```


manateelazycat:
Thanks for patch.

PR #1066, opened 2024-10-08
setting cwd when launching lsp server

fuyu0425:
Some toolchains like rust-analyzer set up by rustup will look rust-toolchain.toml to decide which version to run. 
When there are multiple rust-toolchain.toml across subprojects/dependencies, it's important to sett cwd to the root project path.

### DIFF
--- a/core/lspserver.py
+++ b/core/lspserver.py
@@ -299,11 +299,15 @@ class LspServer:
         self.save_include_text = False
 
         # Start LSP server.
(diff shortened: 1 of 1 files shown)


manateelazycat:
Looks CI is failed, can you test `python -m test.test` in your local termnal ?

manateelazycat:
Thanks, man.

PR #1072, opened 2024-10-12
only re-insert doc when doc change

fuyu0425:
Otherwise, the doc will not be rendered in markdown like the attached gif because of this check.
https://github.com/manateelazycat/lsp-bridge/blob/db75a301714ca203798439bd087897c34dab0f8d/acm/acm.el#L1245
acm-doc-markdown-render-bug


manateelazycat:
Thanks man

PR #1082, opened 2024-10-18
fix expand rust macro

getong:
when using `lsp-bridge-rust-expand-macro` command
error msg
```
Error when processing response 4247
  File "/Users/a/.emacs.d/var/straight/repos/lsp-bridge/core/handler/__init__.py", line 38, in handle_response
    self.process_response(response)
  File "/Users/a/.emacs.d/var/straight/repos/lsp-bridge/core/handler/rust_expand_macro.py", line 15, in process_response
    eval_in_emacs("lsp-bridge-rust-expand-macro--update", response["name"], response["expansion"])
TypeError: 'NoneType' object is not subscriptable
```


manateelazycat:
Thanks

PR #1089, opened 2024-10-24
workspace/symbol should use SYMBOL_MAP instead of KIND_MAP

fuyu0425:
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_symbol


manateelazycat:
Thanks man

PR #1138, opened 2024-12-26
Include `packaging` in pipx dependencies

hzzhang:
`packaging` is imported by core/tabnine.py. Let's include it in the pipx dependencies as well. Otherwise, `pipx run lsp_bridge.py` may complain with errors

```
  File "lsp_bridge.py", line 55, in <module>
    from core.tabnine import TabNine
  File "core/tabnine.py", line 31, in <module>
    from packaging.version import Version
ModuleNotFoundError: No module named 'packaging'
```

### DIFF
--- a/lsp_bridge.py
+++ b/lsp_bridge.py
@@ -5,6 +5,7 @@
 # dependencies = [
 #   "epc",
 #   "orjson",
+#   "packaging",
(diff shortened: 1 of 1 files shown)


manateelazycat:
Thanks

PR #1141, opened 2025-01-01
fix: repair volar windows version and python windows paths

Yuki61803:
1. There was an error in json which caused a parsing error in python.
2. There was a error with using dynamically returned paths for Windows:
SyntaxError: (unicode error) 'unicodeescape' codec can't decode bytes in position 2-3: truncated \UXXXXXXXX escape.
3. There was an error with identifying the LSP server as an object due to the same names of the Volar.

Fix tested on Windows 10.


core/utils.py
```diff
@@ -495,17 +495,17 @@ def replace_template(arg, project_path=None):
         if get_os_name() == "windows":
             user_emacs_dir = get_emacs_func_result("get-user-emacs-directory").replace("/", "\\")
         else:
-            user_emacs_dir = get_emacs_func_result("get-user-emacs-directory")
+            user_emacs_dir = repr(get_emacs_func_result("get-user-emacs-directory")).strip("'")
```
manateelazycat, line 498: If this patch for Windows platform, why repr on non-Windows branch?

lsp-bridge.el
```diff
@@ -452,6 +452,9 @@ Then LSP-Bridge will start by gdb, please send new issue with `*lsp-bridge*' buf
     )
   "The multi lang server rule for file extension."
   :type 'cons)
+(if (string-match "windows" (symbol-name system-type))
+  (setf (cdr (assoc '("vue") lsp-bridge-multi-lang-server-extension-list)) "volar_emmet_windows")
+)
```
manateelazycat, line 457: setf should use in defcustom

core/utils.py
Yuki61803: The path is not processed in any way and python parses the string as unicode and the error that I mentioned above occurs.
SyntaxError: (unicode error) 'unicodeescape' codec can't decode bytes in position 2-3: truncated \UXXXXXXXX escape.

The .replace method above forces the string to be processed without Unicode parsing.

This can be removed, I added it in advance for future code scenarios.
The problem was only in the %USERPROFILE%.

lsp-bridge.el
Yuki61803: I guess it would be better this way

Yuki61803: Can you tell me or write below how to do this right way?

multiserver/volar_emmet_windows.json
```diff
@@ -0,0 +1,22 @@
...
+  "formatting": "volar_windows",
+  "hover": "volar_windows",
+  "signature_help": "volar_windows",
+  "prepare_rename": "volar_windows",
+  "rename": "volar_windows",
+  "document_symbol": "volar_windows",
+  "workspace_symbol": "volar_windows",
+  "semantic_tokens": "volar_windows",
+  "inlay_hint": "volar_windows"
+}
```
manateelazycat, line 22: [volar_windows.json](https://github.com/manateelazycat/lsp-bridge/pull/1141/files#diff-e41316e60c53c2de8309c122307251959ba18c7976ad53e4c1bfd6e3e330cf01) 's name should use "volar",

above also should change to "volar".

lsp-bridge will choose volar_windows.json when it run on Windows platform

langserver/volar_windows.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "volar",
+  "name": "volar_windows",
```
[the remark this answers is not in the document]
Yuki61803: This causes another bug with server not found.
Lsp server gets name volar and program searches it by volar_windows.
So we need to change the name.

multiserver/volar_emmet_windows.json
Yuki61803: This causes another error that I don't remember.
I'll write it tomorrow when i reproduce it.

Yuki61803:
> @Yuki61803 I have push [fb0e319](https://github.com/manateelazycat/lsp-bridge/commit/fb0e31998a2d2589294f0b93c2dfc2bc591222a0)
> 
> You don't need change option lsp-bridge-multi-lang-server-extension-list, lsp-bridge will use volar_emmet_windows.json instead volar_emmet.json on Windows platform.

Ok. I ll check this push tomorrow. My time is 21:33.

Yuki61803:
As you said, everything is correct - it seems I didn't follow the tracing correctly and touched something that time.
In the end, only three lines will change to fix issue with windows volar.

multiserver/volar_emmet_windows.json
Yuki61803: Infinite creation of connections until emacs crashes.

manateelazycat:
@Yuki61803 Thank you very much, good patch! ;)

PR #1144, opened 2025-01-04
Add angular lsp support

Yuki61803:
1. Added two multiservers for editing of angular template files.
There are two types of templates: inline template, external template file).
2. Updated readme for installing angular lsp.
3. Added support for windows version of lsp.
4. Added multilang extensions for angular files.
There is a convention of angular file extensions: https://angular.dev/style-guide (#File structure conventions).
5. Added support for multiple extension levels - previously only one level was available.
Previously extension support:
.html
.ts
Currently extension support:
.component.html
.component.ts
.html
.ts
and etc.

Tested on windows 10.


manateelazycat:
@Yuki61803 Awesome， thank you!

PR #1152, opened 2025-01-21
fix: find last dot position to determine file extension

braineo:
if file name is `abc.spec.ts`, the detected extension is currently `spec.ts` while it should be `ts`


manateelazycat:
I have attach https://github.com/manateelazycat/lsp-bridge/commit/c87e6ac6d452b3bd94583a5c5edb16f37edadb94 try to fix this issue

Thanks for PR

PR #1166, opened 2025-02-21
fix(#1165): use a more reasonable option for log

### LINKED ISSUE #1165: Can we use a more reasonable option for log recording? May be like `lsp-bridge-log-level`?
Currently, the log recording level only have `DEBUG` and `INFO`. But within some lsp server like `rust-analyzer`, the `INFO` level also will output so many log.

Image

Here I just write almost 100 lines of code, but the logs are already very large.

curtainp:
fix #1165

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,6 @@ and add this to your `config.el`
 ``` elisp
 (use-package! lsp-bridge
   :config
-  (setq lsp-bridge-enable-log nil)
   (global-lsp-bridge-mode))
 ```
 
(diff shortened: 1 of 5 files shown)


manateelazycat:
This patch cannot pass CI， below is CI error:

SyntaxError: invalid syntax
Process *lsp-bridge* exited abnormally with code 1
waiting for remote process to start...
  File "/home/runner/work/lsp-bridge/lsp-bridge/lsp_bridge.py", line 174
    match str(lsp_server_log_level):
          ^

manateelazycat:
use if else instead match？

manateelazycat:
Thanks

PR #1172, opened 2025-03-05
Add customizable imenu function

Y0ngg4n:
This add the functionality to also use the imenu-consult function for a better imenu


manateelazycat:
Thanks for patch

Issue #1183, opened 2025-03-27
Copilot Support: Considering the Official SDK


mopemope:
Hello,

Regarding lsp-bridge's Copilot support, are there any plans to modify it to use `@github/copilot-language-server`?

I believe it would be better to use the official SDK, as it is available.

Could I get your thoughts on this?

manateelazycat:
PR are welcome!

PR #1185, opened 2025-03-28
feat: switch to use the official @github/copilot-language-server for copilot support.

mopemope:
Switched the server used for Copilot support to the official @github/copilot-language-server. 
This enables the use of features like `Inline Completions` in the future.

#1183


manateelazycat:
@mopemope Thanks man. ;)

PR #1187, opened 2025-03-28
feat: implement inline completion support for copilot

mopemope:
Implemented inline completions.
Modified to call the `textDocument/inlineCompletion` API during completion, referring to https://github.com/github/copilot.vim/.

Reference:
https://github.com/github/copilot-language-server-release?tab=readme-ov-file#inline-completions

### DIFF
--- a/core/copilot.py
+++ b/core/copilot.py
@@ -1,6 +1,7 @@
 import os
 import time
 import subprocess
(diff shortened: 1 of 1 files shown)


manateelazycat:
Awesome!

PR #1188, opened 2025-03-28
feat: add workspace folder support to copilot.

mopemope:
This change modifies the code to provide the workspace folder information to Copilot, aiming to enhance the accuracy of Copilot's code completion suggestions.

PR #1190, opened 2025-03-30
fix: don't start copilot when not used

kira-bruneau:
#1188 introduced a change that unintentionally starts copilot, even if you don't use it. If you didn't have nodejs installed, this would break lsp-bridge.


MatthewZMD:
Great, thanks!

PR #1192, opened 2025-04-16
Fixed epc:ReturnError when editing remote files.

i-aki-y:
In the current master, completion would fail in remote editing by ssh.
This is an example of the log:
```
ERROR:epc:ReturnError([Symbol('invalid-function'), ['(lsp-bridge--record-work-done-progress \'"[LSP-Bridge] ca60a9bf-6cba-4922-b869-257998df7c5a 1 file to analyze")']])
````

I found this is caused by misuse of `eval_sexp_in_emacs(message["sexp"])` in `lsp_bridge.py` where `eval_sexp_in_emacs` expect a string argument but the `message["sexp"]` appears to be a list ( = `[sexp]`). This causes passing a list of a list in the `epc_client.call()` function, see also:

```py
def eval_sexp_in_emacs(sexp):
    epc_client.call("eval-in-emacs", [sexp]). #  since the `sexp` is already a list, the `[sexp]` would be a list of a list
```

This PR would fix this bug. (also fix a typo `_get_remtoe_file_client `)


i-aki-y:
> TypeError: 'type' object is not subscriptable

The `list[str] ` style typing is supported in pythion > 3.8, but the CI looks to use python3.8.
However, the python3.8 is no longer supported:  https://devguide.python.org/versions/

Are you planning to upgrade the python version of CI?

manateelazycat:
I have upgraded the Python in CI to 3.13, but CI is still reporting an error. Could you take a look at it again? Thank you for your patch.

i-aki-y:
@manateelazycat Thanks! I guess old yaml was referred, I merged new master and now looks fine.

manateelazycat:
@i-aki-y Thanks a lot for your patch, it's great!

PR #1201, opened 2025-05-15
Remove unnecessary `(interactive)`

zonuexe:


lsp-bridge-ref.el
```diff
@@ -363,7 +362,6 @@ Becomes buffer local in `lsp-bridge-ref-mode' buffers.")
 (defun lsp-bridge-ref-popup (references-content references-counter)
   "Rerun rg with customized arguments. This function will give
 user more freedom to use rg with special arguments."
-  (interactive)
```
zonuexe, line 366: This function is not called as a command. It is listed as a suggestion for `M-x`, so it is noise.

PR #1202, opened 2025-05-20
acm-backend-copilot-show-trailing: mimic VSCode

fuyu0425:
I finally just upgraded lsp-bridge to the latest version again yesterday lol.

The new integration of copilot-language-server with inlineCompletion will show the whole line as displayLabel, which is not the same as VSCode and the old version of lsp-bridge with copilot-node-server: which only show the trailing parts of the completions.

If the current line is ""a b cd" and copilot returns ""a b cde fg"; the displayLabel will be changed to "cde fg": keeps the current words and the trailings.

I don't know where to put the change; so just put it in the `acm.el` in the `acm-menu-candidates-preprocess` by a calling a new preprocessing function `acm-menu-candidates-preprocess`.
To get the current line in the original buffer (instead of the overlay buffer), I use `lsp-bridge--last-buffer`, looks a little strange... (any better way to do that??) or we should do it in `copliot.py` (still need to get the current line content though)

I added an option `acm-backend-copilot-show-trailing` to enable this behavior; default to false though, you can decide whether to make it the default (which I prefer).

### DIFF
--- a/acm/acm-backend-copilot.el
+++ b/acm/acm-backend-copilot.el
@@ -29,6 +29,10 @@
   :type 'string
   :group 'acm-backend-copilot)
 
+(defcustom acm-backend-copilot-show-trailing nil
+  "Show only the trailing parts for copilot. (like the VSCode)"
+  :group 'acm-backend-copilot)
+
 (defcustom acm-backend-copilot-accept nil
   "Send accept request."
   :type 'boolean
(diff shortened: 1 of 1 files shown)


acm/acm.el
```diff
@@ -1044,7 +1044,34 @@ The key of candidate will change between two LSP results."
...
+                             (prefix-len (length prefix))
+                             (trailing-text (if (string-prefix-p prefix copilot-line)
+                                                (substring copilot-line prefix-len)
+                                              copilot-line)))
+                        (plist-put candidate :displayLabel trailing-text)))
+                    candidate))
+                acm-menu-candidates)))
+
 (defun acm-menu-render (menu-old-cache)
+  (acm-menu-candidates-preprocess)
```
manateelazycat, line 1074: Should use (string-equal backend "copilot") wrap here.

Otherwise, all backend menu item will did check, it will slow down acm

manateelazycat (changes requested):

acm/acm.el
fuyu0425: I thought `acm-menu-candidates` is the merged results of all backends?  Isn't it?
How do I get the backend here without iterating `acm-menu-candidates`??

fuyu0425: I will try to do that in 
```
    (when acm-enable-copilot
      (setq copilot-candidates (acm-backend-copilot-candidates keyword)))
```

fuyu0425:
I move all changes  in acm-backend-copilot.el
looks better now.

manateelazycat:
Looks good, thanks

Issue #769, opened 2023-11-09
Connection issues with lsp-bridge-enable-with-tramp and doom emacs


Bawaw:
Hey, thank you for this package, it is the best experience I've had with LSP in emacs so far!

However, while trying to integrate lsp-bridge in my daily workflow (doom emacs + remote pyright) I hit a block. When I `(setq lsp-bridge-enable-with-tramp 't)` I fail to connect to the LSP server when opening a file using tramp (see logs below). 

PS: When opening a remote file using `lsp-bridge-open-remote-file` everything works as expected. 

Details: 
- Ubuntu 22.04.3 LTS
- GNU Emacs 29.1
- Doom core     v3.0.0-pre
- Doom modules  v23.09.0-pre

```
Eval in Emacs: (lsp-bridge--first-start '60257)
Traceback (most recent call last):
  File "/home/bawaw/.lspbridge/lsp-bridge/lsp_bridge.py", line 180, in send_message_dispatcher
    client = self.get_socket_client(data["host"], port)
  File "/home/bawaw/.lspbridge/lsp-bridge/lsp_bridge.py", line 398, in get_socket_client
    self.host_names[server_host]["username"],
KeyError: 'remotehostname'

Traceback (most recent call last):
  File "/home/bawaw/.lspbridge/lsp-bridge/core/remote_file.py", line 63, in connect_ssh
    ssh.connect(self.ssh_host, port=self.ssh_port, username=self.ssh_user, key_filename=pub_key)
  File "/home/bawaw/.local/lib/python3.10/site-packages/paramiko/client.py", line 377, in connect
    to_try = list(self._families_and_addresses(hostname, port))
  File "/home/bawaw/.local/lib/python3.10/site-packages/paramiko/client.py", line 202, in _families_and_addresses
    addrinfos = socket.getaddrinfo(
  File "/usr/lib/python3.10/socket.py", line 955, in getaddrinfo
    for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
OSError: Int or String expected

Traceback (most recent call last):
  File "/home/bawaw/.lspbridge/lsp-bridge/core/remote_file.py", line 63, in connect_ssh
    ssh.connect(self.ssh_host, port=self.ssh_port, username=self.ssh_user, key_filename=pub_key)
  File "/home/bawaw/.local/lib/python3.10/site-packages/paramiko/client.py", line 377, in connect
    to_try = list(self._families_and_addresses(hostname, port))
[...]

PR #1212, opened 2025-06-26
Enhance Remote SSH experience

fuyu0425:
I finally tested lsp-bridge with tramp (to compare with ssh + tmux + TUI emacs).

I have added some enhancements to fit my needs.

## 1. Allow the remote server and the local server to coexist. 
To differentiate them, I test whether it specified a port number when launching `lsp_bridge.py`; I also change to use `pgrep` instead of the original `ps + grep + awk`

## 2. Using SSH-Agent
Some of my key is encrypted, and I see the `Key is encrypted` errors like #769 #939 
I add all my keys (encrypted or unencrypted) to ssh-agent with Keychain (https://www.funtoo.org/Funtoo:Keychain), which manages the keys and adds them to ssh-agent.

I added a new option `lsp-bridge-user-ssh-agent` (default to false, so existing users will not notice the difference).
If turned on, it will not specify `key_filename` and paramiko will try to use keys in ssh-agent.

## 3. Resolve non-IP hostname with DNS
I tried to rework it to allow non-IP hostnames, but failed lol.
So, I used an easier way to resolve the IP with DNS and cache the result.

I test and it works with my ssh_config.
```
Host work
    Hostname work.home
    User yufu
    ForwardX11 yes
    ForwardX11Trusted yes
```
with tramp file path `/sshx:work:/home/yufu/x.py`

### DIFF
--- a/core/remote_file.py
+++ b/core/remote_file.py
@@ -52,7 +52,9 @@ class RemoteFileClient(threading.Thread):
         self.callback = callback
         [self.remote_python_command, self.remote_python_file, self.remote_log] = get_emacs_vars(["lsp-bridge-remote-python-command", "lsp-bridge-remote-python-file", "lsp-bridge-remote-log"])
 
-        [self.user_ssh_private_key] = get_emacs_vars(["lsp-bridge-user-ssh-private-key"])
+        [self.user_ssh_private_key,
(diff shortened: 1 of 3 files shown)


manateelazycat:
Cool man

PR #1213, opened 2025-06-27
uv integration

fuyu0425:
- uv integration
- `pyhon-lsp-bridge` uv wrapper
   - used by default if in path 	
-  README


manateelazycat:
Thanks

PR #1228, opened 2025-08-02
Add `lsp-bridge-breadcrumb-mode`

deirn:
Screenshot From 2025-08-02 10-50-10

Is there a way to know LSP server capabilities in Emacs side? Would've been better to auto hide if server doesn't support [`documentSymbol`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#documentSymbolOptions).


manateelazycat:
@deirn This feature is amazing, thanks for your work.

I see there's a CI error. Could you take another look at the patch? I'll merge this patch after CI checks it.

Thanks again for your work.

manateelazycat:
thanks,man

PR #1237, opened 2025-08-13
Breadcrumb: collapse parent segments if window width is too small

deirn:
Also add max segment character length option.

image
image

### DIFF
--- a/lsp-bridge-breadcrumb.el
+++ b/lsp-bridge-breadcrumb.el
@@ -52,6 +52,11 @@
   :type 'string
   :group 'lsp-bridge-breadcrumb)
 
+(defcustom lsp-bridge-breadcrumb-max-segment-length 20
+  "Max character length for each segment."
+  :type 'integer
+  :group 'lsp-bridge-breadcrumb)
+
 (defvar-local lsp-bridge-breadcrumb--timer nil)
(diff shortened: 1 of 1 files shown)


deirn:
`lsp-bridge.el` doesn't need `lsp-bridge-breadcrumb.el` to function.

Czcfsdx:
True. But if we don't add `(require 'lsp-bridge-breadcrumb)` in `lsp-bridge.el`, it will report error `lsp-bridge/lsp-bridge.el failed to define function lsp-bridge-breadcrumb-mode`  and we couldn't find the command `lsp-bridge-breadcrumb-mode`

Like `lsp-bridge-semantic-tokens.el` or `lsp-bridge-rust.el`, `lsp-bridge.el` doesn't need them to function, but they add `(require)` in `lsp-bridge.el` to provide their function

It is OK to let users add `(require 'lsp-bridge-breadcrumb)` in their configuration, but we should tell users in README

deirn:
That's interesting, maybe I don't encounter this issue since I use elpaca instead of loading lsp-bridge manually.

Czcfsdx:
I load lsp-bridge manually, don't know that elpaca can handle this automatically

PR #1243, opened 2025-08-29
Improve doc frame positioning and add boundary constraint configuration

falood:
This PR fixes a bug where the documentation frame would overflow beyond the bottom screen edge and adds a configuration option for boundary constraints.

## Changes:
Fixed bottom overflow bug: Documentation frame no longer gets cut off when positioned near the bottom of the screen. It now automatically repositions upward when there's insufficient space below
Enhanced space utilization: Documentation frame can now use up to the full Emacs frame size while respecting display boundaries
Configurable boundary constraint: Added acm-doc-frame-boundary option allowing users to choose between:
* 'parent-frame (default): Constrain doc frame within Emacs window
* 'display: Allow doc frame to extend beyond Emacs window but within display screen

### bottom screen edge (Before)

telegram-cloud-document-5-6165932170570372739

### bottom screen edge (Fixed)

telegram-cloud-photo-size-5-6165932171026615016-y

### doc frame size (default, acm-doc-frame-boundary: parent-frame)

8566fa67c2c10f6f180769e395f85233

### doc frame size (acm-doc-frame-boundary: display)

telegram-cloud-photo-size-5-6165932171026615021-y

### DIFF
--- a/README.md
+++ b/README.md
@@ -332,6 +332,7 @@ lsp-bridge provides support for more than two language servers for many language
 - `acm-enable-citre`: Integration with [citre(ctags)](https://github.com/universal-ctags/citre). Enable this to add citre (ctags) backend (disabled by default)
 - `acm-enable-lsp-workspace-symbol`: Enable LSP workspace symbol completion, disabled by default
 - `acm-doc-frame-max-lines`: Max line number of help documentation, default is 20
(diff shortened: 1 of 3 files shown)


manateelazycat:
Thanks

PR #1249, opened 2025-09-17
add move init option

getong:
see https://github.com/MystenLabs/sui/blob/eef60efd56b80cabf3c808602723988a24995c4a/external-crates/move/crates/move-analyzer/src/analyzer.rs#L187C5-L208C7

There are three init options here, so enable them in the json file.


manateelazycat:
Thanks

PR #1251, opened 2025-09-18
make sure only one process started

deirn:
Since `lsp-bridge-epc-process` is set only after the Python server called `lsp-bridge--first-start`, `lsp-bridge-process-live-p` can return nil even if the Emacs-side process is already started.

This introduces race condition when switching to new buffer too quickly before the server calls first-start, making the server started again.

### DIFF
--- a/lsp-bridge.el
+++ b/lsp-bridge.el
@@ -1317,7 +1317,8 @@ So we build this macro to restore postion after code format."
 
 (defun lsp-bridge-start-process ()
   "Start LSP-Bridge process if it isn't started."
-  (if (lsp-bridge-process-live-p)
+  (if (and (process-live-p lsp-bridge-server)
+           (process-live-p lsp-bridge-internal-process))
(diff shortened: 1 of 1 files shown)


deirn:
It's not because this PR

manateelazycat:
> Figured it, see #1255

Sorry, I just push https://github.com/manateelazycat/lsp-bridge/commit/2e795d54fc99f4d238180347424b04bd33e0c0c5 to revert this patch

you mean CI failed is because 1255 ?

deirn:
It's because one of the test dependency (Tempel) updated to no longer supports the test Emacs version, #1255 fixed this

manateelazycat:
@deirn I see, I have repush your patch, thanks a lot

https://github.com/manateelazycat/lsp-bridge/commit/c678b11a030af2c3fc786463f0675cb963231e69

Sorry again

PR #1260, opened 2025-11-09
Update python-lsp-bridge shebang line

Yucklys:
### Summary

The original shebang line for `python-lsp-bridge` is `#!/bin/bash`, which does not exist for distro like NixOS. This will cause `No such file or directory` error on any distro that doesn't put bash on `/bin/bash`. For portability concern, it is better to use `#!/usr/bin/env bash` instead.

### Changes
- Change shebang line of `python-lsp-bridge` to `/usr/bin/env bash`


manateelazycat:
Thanks for patch! Cool

Issue #1256, opened 2025-10-25
build errors with python 3.14


thromer:
uv by default will use python3.14, but two of lsp-bridge's dependencies don't build out of the box on debian bookworm / x86_64 (paramiko and orjson). Setting `requires-python = "==3.13.*"` in `pyproject.toml` is the quickest fix for me.

It also works to instead update orjson to `>=3.11.4` and pynacl to `>=1.6.0` as suggested on this [paramiko issue](https://github.com/paramiko/paramiko/issues/2550), though IMO it would be ugly for lsp-bridge to reference pynacl as a direct dependency.

PR #1264, opened 2025-12-11
feat: constrain python version to exclude 3.14

nailuoGG:
Update requires-python constraint from >=3.13 to >=3.13,<3.14 to prevent compatibility issues with Python 3.14.

#1256 

A minor change that helps keep the project working out of the box.


manateelazycat:
Thanks for patch

PR #1269, opened 2025-12-29
Add default option to multiserver config

deirn:
When an option is missing in a multiserver config JSON, use the `default` option.


manateelazycat:
@deirn Cool!

PR #1272, opened 2026-01-13
Fix error when evaluating buffer directly

zonuexe:
Handle cases where `load-file-name` is nil by falling back to `default-directory`.  This allows the code to be evaluated directly without errors.

### DIFF
--- a/acm/acm-icon.el
+++ b/acm/acm-icon.el
@@ -173,7 +173,9 @@
     (t . ("material" "smoking-pipe" "#90cef1"))))
 
 (defvar acm-icon-cache (make-hash-table :test 'equal))
-(defvar acm-icon-dir (expand-file-name "icons" (file-name-directory load-file-name)))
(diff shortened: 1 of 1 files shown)


zonuexe:
This is a common pattern in Emacs Lisp to support both loading and direct evaluation. Similar approaches are used in other projects as well:

 - https://github.com/abo-abo/swiper/blob/ec9421340c88ebe08f05680e22308ed57ed68a3d/ivy.el#L5493-L5501
 - https://github.com/manateelazycat/holo-layer/blob/dc7d48338b7c243fe5660be40cedb5af9e0c94c5/holo-layer.el#L88-L90

After submitting this PR, I realized that the following might be a more concise way to handle it:

```el
(defvar acm-icon-dir (expand-file-name "icons" (file-name-directory (or load-file-name buffer-file-name))))
```

Since this is not a critical issue, I don't intend to pursue this specific change further or update the PR. However, I remain happy to contribute to the project's development in other areas. Thanks!

PR #1294, opened 2026-04-11
Add custom variable `lsp-bridge-ref-popup-window-handler`

Czcfsdx:
Add custom variable `lsp-bridge-ref-popup-window-handler` for customizing how to creating and opening a new lsp-bridge-ref window.

Move the origin behavior to `lsp-bridge-ref-popup-window-default-handler` as the default behavior.


manateelazycat:
Thanks for patch

PR #1299, opened 2026-06-07
Show LSP initialization progress in mode-line & Improve cursor positioning for type inlay hints

### LINKED ISSUE #1298: Show $/progress loading status in modeline during rust-analyzer initialization
When opening a Rust file in a large project (e.g. Bevy), rust-analyzer goes through several initialization stages before it is ready to serve completions and diagnostics. During this time, lsp-bridge gives no visual feedback, so the user has no way to tell whether the LSP is still loading or has already failed silently.
What rust-analyzer sends
Throughout initialization, rust-analyzer emits a series of $/progress notifications with begin / report / end lifecycle events. The key tokens, are:

rustAnalyzer/Fetching
rustAnalyzer/Building CrateGraph
rustAnalyzer/Loading proc-macros
rustAnalyzer/Roots Scanned (carries a message field like "230/577")
rustAnalyzer/Indexing

This is already documented in the LSP spec as window/workDoneProgress, and lsp-bridge's log shows it is already receiving these notifications (https://github.com/manateelazycat/lsp-bridge/issues/1100).

**Expected behavior**
Between the begin and end of these progress tokens, the modeline should show a loading indicator such as:
[rust-analyzer load-icon 230/577]
Once the final stage ends, the indicator should be cleared. This matches the behavior VS Code and lsp-mode(https://github.com/emacs-lsp/lsp-mode/pull/707) provides out of the box — it shows a spinner in the status bar with the current stage name and progress during the entire initialization window.

Image
(the picture is token from here: https://www.youtube.com/watch?v=jq-zvM4Ih14)

not-null-nil:
Close #1298: Show LSP initialization progress in mode-line

## What this PR does

During rust-analyzer initialization, lsp-bridge now displays real-time progress in the mode-line of affected buffers, so users can tell whether the LSP is still loading rather than wondering if it has failed silently.

Try to improve cursor positioning for type inlay hints

## Changes

**`core/lspserver.py`**

- Pass `file_paths` (the list of files currently open in the server) as an additional argument to `lsp-bridge--record-work-done-progress`, so the progress indicator is only shown in buffers that belong to the server currently loading

**`lsp-bridge.el`**

- New buffer-local variable `lsp-bridge--work-done-progress` and `lsp-bridge--work-done-progress-timer` so each buffer tracks its own progress state independently
- Update `lsp-bridge--record-work-done-progress` to iterate over `file-paths` and update only the matching buffers, with a 1-second auto-clear timer per buffer
- Add `lsp-bridge--format-progress-message` to format the raw progress string: extracts stage name and percentage, strips the `[LSP-Bridge]` prefix and `rustAnalyzer/` namespace, and shows a nerd-icons icon when available
- Update `lsp-bridge--mode-line-format` to render the buffer-local progress string alongside the existing `lsp-bridge` label

Clears automatically once loading is complete.

## Notes

- Works for any LSP server that emits `$/progress` notifications, not just rust-analyzer
- Progress is shown only in buffers belonging to the loading server, not globally
- The existing `acm-backend-lsp-show-progress` behavior is preserved

## Try to improve cursor positioning for type inlay hints

When moving the cursor rightward through a line with type inlay hints (InlayHintKind=1), the cursor previously jumped past the entire hint text, which was visually jarring. This is a known issue with Emacs `before-string' overlays (ref: emacs-lsp/lsp-mode#3263).

The fix applies the `cursor 1' text property to all type hint overlays, so the cursor stops at the beginning of the hint instead of skipping over it.

### DIFF
--- a/core/lspserver.py
+++ b/core/lspserver.py
@@ -799,7 +799,7 @@ class LspServer:
             ("rename_prepare_provider", ["result", "capabilities", "renameProvider", "prepareProvider"]),
             ("code_action_provider", ["result", "capabilities", "codeActionProvider"]),
             ("code_action_kinds", ["result", "capabilities", "codeActionProvider", "codeActionKinds"]),
-            ("document_highlight_provider", ["result", "capabilities", "documentHighlightProvider"]),            
+            ("document_highlight_provider", ["result", "capabilities", "documentHighlightProvider"]),
             ("code_format_provider", ["result", "capabilities", "documentFormattingProvider"]),
             ("range_format_provider", ["result", "capabilities", "documentRangeFormattingProvider"]),
             ("signature_help_provider", ["result", "capabilities", "signatureHelpProvider"]),
@@ -905,7 +905,8 @@ class LspServer:
                     progress_message += str(message_attr)
 
             if progress_message != "":
-                eval_in_emacs("lsp-bridge--record-work-done-progress", "[LSP-Bridge] " + progress_message)
+                file_paths = list(self.files.keys())
+                eval_in_emacs("lsp-bridge--record-work-done-progress", "[LSP-Bridge] " + progress_message, file_paths)
 
     def handle_register_capability_message(self, message):
         if "method" in message and message["method"] in ["client/registerCapability"]:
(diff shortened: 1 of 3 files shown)


manateelazycat:
Thanks for patch

PR #1300, opened 2026-06-12
fix: reply to unhandled server->client requests (gopls session hang)

puluo2void:
## Problem

gopls sends a `client/unregisterCapability` request when it re-registers its file watchers. This happens ~0.5s after `didOpen` whenever the workspace root contains **more than one Go module** (e.g. a repo root with `backend/go.mod`, `tools/go.mod`, …): gopls discovers the modules, recalculates views, registers the new watcher set, then unregisters the old one.

lsp-bridge never responds to that request, and gopls blocks its session waiting for the reply. From that moment **every** request on the session — definition, references, hover, signatureHelp, formatting — hangs forever with no error. From the user's side, all LSP commands silently do nothing, and restarting lsp-bridge re-wedges within a second because the register/unregister cycle repeats on every fresh session.

## Root cause

`handle_recv_message` answers exactly four server→client request methods (`workspace/configuration`, `workspace/applyEdit`, `window/workDoneProgress/create`, `client/registerCapability`). Any other server request — `client/unregisterCapability`, `window/showMessageRequest`, `workspace/workspaceFolders` — is silently dropped. The LSP spec requires the client to respond to every server request; gopls blocks on the missing reply, and other servers may too.

## Reproduction (no Emacs involved)

A minimal stdio LSP client that mirrors lsp-bridge's capabilities and reply behavior, with gopls v0.22.0, rooted at a workspace containing two Go modules:

| gopls mode | reply to `unregisterCapability` | result |
|---|---|---|
| `gopls` (plain) | withheld (current lsp-bridge behavior) | **wedged** — zero further messages from gopls; `textDocument/definition` never answered |
| `gopls -remote=auto` | withheld | **wedged** identically |
| either | replied `null` | works — definition answered in ~50ms |

<details><summary>Repro script (adjust PROJ/FILE paths to a multi-module Go workspace)</summary>

```python
#!/usr/bin/env python3
"""Usage: gopls_wedge_test.py {plain|daemon} {reply|withhold}"""
import json, subprocess, sys, threading, time

MODE, BEHAVIOR = sys.argv[1], sys.argv[2]
PROJ = "/path/to/multi-module-workspace"      # repo root containing >1 go.mod
FILE = PROJ + "/backend/internal/some/file.go"  # any .go file in a sub-module
ROOT_URI, FILE_URI = "file://" + PROJ, "file://" + FILE

cmd = ["gopls"] if MODE == "plain" else ["gopls", "-remote=auto"]
p = subprocess.Popen(cmd, cwd=PROJ, stdin=subprocess.PIPE,
                     stdout=subprocess.PIPE, stderr=subprocess.DEVNULL)
lock, events, responses = threading.Lock(), [], {}
unregister_seen, init_done = threading.Event(), threading.Event()
t0 = time.time()

def log(d, s):
    with lock: events.append((time.time() - t0, d, s))

def send(msg):
    msg.setdefault("jsonrpc", "2.0")
    data = json.dumps(msg).encode()
    with lock:
        p.stdin.write(b"Content-Length: %d\r\n\r\n%b" % (len(data), data))
        p.stdin.flush()
    log(">>", msg.get("method") or "response(id=%s)" % msg.get("id"))

def reader():
    while True:
        line = p.stdout.readline()
        if not line:
            return log("!!", "gopls stdout closed")
        if not line.lower().startswith(b"content-length:"):
            continue
        n = int(line.split(b":")[1])
        while p.stdout.readline() not in (b"\r\n", b"\n"): pass
        msg = json.loads(p.stdout.read(n))
        mid, method = msg.get("id"), msg.get("method")
        if method and mid is not None:
            log("<<", "request %s (%s)" % (method, mid))
            if method == "client/unregisterCapability":
                unregister_seen.set()
                if BEHAVIOR == "withhold":
                    log("--", "WITHHOLDING reply to %s" % mid); continue
                send({"id": mid, "result": None})
            elif method == "workspace/configuration":
                send({"id": mid, "result": [None] * len(msg["params"]["items"])})
            else:
                send({"id": mid, "result": None})
        elif method:
            log("<<", "notification %s" % method)
        else:
            log("<<", "response (%s)" % mid)
            responses[mid] = msg
            if mid == 1: init_done.set()

threading.Thread(target=reader, daemon=True).start()
send({"id": 1, "method": "initialize", "params": {
    "processId": None, "rootUri": ROOT_URI, "clientInfo": {"name": "wedge-test"},
    "capabilities": {"workspace": {
        "configuration": True, "workspaceFolders": True,
        "didChangeWatchedFiles": {"dynamicRegistration": True,
                                  "relativePatternSupport": True}}}})
assert init_done.wait(20), "initialize never answered"
send({"method": "initialized", "params": {}})
send({"method": "workspace/didChangeConfiguration", "params": {"settings": {
    "gopls": {"usePlaceholders": True, "completeUnimported": True,
              "staticcheck": True}}}})
text = open(FILE).read()
send({"method": "textDocument/didOpen", "params": {"textDocument": {
    "uri": FILE_URI, "languageId": "go", "version": 1, "text": text}}})
line_no, char_no = next((i, l.index(l.split()[0])) for i, l in
                        enumerate(text.splitlines()) if "func " in l)
got = unregister_seen.wait(25); time.sleep(1)
send({"id": 99, "method": "textDocument/definition", "params": {
    "textDocument": {"uri": FILE_URI},
    "position": {"line": line_no, "character": char_no}}})
deadline = time.time() + 12
while time.time() < deadline and 99 not in responses: time.sleep(0.2)
print("unregister received: %s | definition answered: %s"
      % (got, 99 in responses))
for t, d, s in events[-12:]: print("  %6.2fs %s %s" % (t, d, s))
p.kill()
```
</details>

This matches what users see in the `*lsp-bridge*` log: a `Recv client/unregisterCapability request (N)` line with no `Send response to server request N` after it, then no `Recv` lines ever again while `Send` lines keep accumulating.

## Fix

1. Answer `client/unregisterCapability` with a `null` result (mirroring the existing `client/registerCapability` handling).
2. Add a catch-all in `handle_recv_message`: any server→client *request* that no existing handler answered gets a `null` reply plus a log line. `null` is a legal response for the remaining methods (`window/showMessageRequest` → "user dismissed the prompt", `workspace/workspaceFolders` → `null` is allowed by the spec), and never replying deadlocks servers that block on their requests.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/core/lspserver.py
+++ b/core/lspserver.py
@@ -924,6 +924,31 @@ class LspServer:
 
             self.sender.send_response(message["id"], None)
 
+        if "method" in message and message["method"] in ["client/unregisterCapability"]:
+            # Reply is mandatory: gopls blocks its session (and the shared
+            # daemon forwarder) waiting for this response, after which every
+            # request (definition, references, etc.) hangs forever.
+            self.sender.send_response(message["id"], None)
+
+    ANSWERED_SERVER_REQUEST_METHODS = [
+        "workspace/configuration",
+        "workspace/applyEdit",
+        "window/workDoneProgress/create",
+        "client/registerCapability",
+        "client/unregisterCapability",
+    ]
+
+    def handle_unanswered_request_message(self, message):
+        # Every server->client *request* needs a response, or servers like
+        # gopls block their whole session on the missing reply. Null is a
+        # legal response for the requests we can't act on (e.g.
+        # window/showMessageRequest means "user dismissed the prompt").
+        if "id" in message and "method" in message and \
+           message["method"] not in self.ANSWERED_SERVER_REQUEST_METHODS:
+            log_time("Send null response to unhandled server request {} ({})".format(
+                message["method"], message["id"]))
+            self.sender.send_response(message["id"], None)
+
     def handle_recv_message(self, message: dict):
         if "error" in message:
             self.handle_error_message(message)


manateelazycat:
Thanks for help

PR #1301, opened 2026-06-26
fix: improve remote connection recovery after disconnect

fzcsyglq:
1. When ~/.ssh/config is used, Host and IP mismatch during reconnection, preventing channel recovery. Fixed by reverse look-up in SSH config.
2. LSP servers were killed on channel disconnect but not restarted after reconnect.  Fixed by removing the kill on disconnect. 
3. Fix some potential deadlock on reconnect.


manateelazycat:
Thanks for patch

PR #1309, opened 2026-08-20
Add new JavaScript and TypeScript options

jrrom:
I've tried my best to add these options so that LSPs for JavaScript and TypeScript can be handled much more easily in the future.

Sorry if there are any errors.

Thank you


manateelazycat:
Thanks bro

PR #1311, opened 2026-08-25
fix: Add c-or-c++-mode to C/C++ lang server mode list for .h files

fzcsyglq:
Fix .h files failing to find clangd by adding c-or-c++-mode to the mode list.


manateelazycat:
Thanks

PR #1312, opened 2026-09-16
Fix inline Org Babel completion and missing-server crash

### LINKED ISSUE #1308: org mode completetion
When open a org file, the lsp-bridge reports the error:
```
  File "/home/lqz/.config/emacs/.local/straight/build-30.1/lsp-bridge/lsp_bridge.py", line 714, in event_dispatcher
    getattr(self, func_name)(*func_args)
  File "/home/lqz/.config/emacs/.local/straight/build-30.1/lsp-bridge/lsp_bridge.py", line 1009, in _do
    open_file_success = self.open_file(filepath)  # _do is called inside event_loop, so we can block here.
                        ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lqz/.config/emacs/.local/straight/build-30.1/lsp-bridge/lsp_bridge.py", line 751, in open_file
    lang_server_info = load_single_server_info(single_lang_server)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lqz/.config/emacs/.local/straight/build-30.1/lsp-bridge/lsp_bridge.py", line 1124, in load_single_server_info
    with open(lang_server_info_path, encoding="utf-8", errors="ignore") as f:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/home/lqz/.config/emacs/.local/straight/repos/lsp-bridge/langserver/[].json'
```
it works perfect for other files.  How can  i fix it?

nujabse:
## Summary

Fixes #1308.

Restore inline Org Babel server lookup while keeping the `C-c '` source-edit
buffer support introduced in #1303. Also avoid loading a language-server
configuration when an Org request has no valid server name.

## Cause

Commit 8da65749869ac1f99efbde521441edb51bdbe1bd made the `.org` branch of
`lsp-bridge-get-single-lang-server-by-file-mode` look exclusively for a
separate `org-src-mode` buffer. If the user edits a block directly in the Org
buffer, that branch returns nil instead of reaching the original Babel lookup.

EPC represents nil as `[]` on the Python side. `LspBridge.open_file` passes it
unconditionally to `load_single_server_info`, causing
`FileNotFoundError: .../langserver/[].json`. The exception escapes the event
dispatch loop, so restarting without fixing lookup does not resolve the cause.

## Reproduction

1. Enable `lsp-bridge-enable-org-babel` and use a working Python language server.
2. Open an Org file containing the following block and put point in its body.
   Do not open a separate `C-c '` buffer.

   ```org
   #+begin_src python
   import os
   #+end_src
   ```

3. Compare `(lsp-bridge-org-babel-check-lsp-server)` with
   `(lsp-bridge-get-single-lang-server-by-file-mode buffer-file-name)`.
   Before the fix, the former returns the configured server and the latter nil.
4. Trigger completion: the Python backend can attempt to load `[].json`.

This also affects `jupyter-python` when mapped to `python`; Jupyter itself is
not required for the reproduction.

## Changes

- When no source-edit buffer exists, look up the server in the matching Org
  buffer, not in an arbitrary EPC callback buffer.
- Preserve source-edit buffer precedence and the Wen/ds-pinyin alternatives.
- Return False before loading Org server configuration if the name is empty
  or not a string. This also handles point leaving a block before a queued
  request is processed.
- Add ERT and Python regression tests; run the ERT tests from the existing
  test entrypoint before starting the bridge.

## Validation

- Against the unmodified upstream lookup function, the same ERT suite fails
  4 tests (inline Python, inline Jupyter Python, callback context, Org alternatives).
  All 9 pass with this patch.
- 9 ERT tests passed: inline Python/Jupyter Python, unrelated callback buffer,
  source-edit precedence, block boundaries, disabled/excluded Babel,
  Wen/ds-pinyin alternatives, missing Org buffer, and ordinary Python files.
- `python3 -m unittest test.test_org_babel -v`: 2 tests passed, including
  empty-list/None/empty-string values and a valid server.
- Live Emacs 30.2.50/basedpyright validation returned 728 import candidates
  and recorded 100 in the Emacs completion cache after applying the fix.
- Full upstream integration suite not run locally.

The separate Jupyter error caused by an empty `:session` header is unrelated
and intentionally not changed by this PR.


manateelazycat:
Thanks

PR #1313, opened 2026-09-17
Add PowerShell Editor Services support

lazyskulptor:
## Summary
- map `powershell-mode` to PowerShell Editor Services and enable its default hook
- add a stdio language-server definition with configurable environment paths
- expand arbitrary `$VAR` command templates and cover the configuration with tests

## Test plan
- `python3 -m test.test`


manateelazycat:
Thanks

